PR #4, opened 2016-11-20
Adds --midi-learn option in the manpage

trebmuh:
Hope this is right and it helps. You might want to tweak a better definition maybe.


fundamental:
Thanks

trebmuh:
:+1:

PR #7, opened 2017-05-08
Make plugin respond to midi program change

dsheeler:


fundamental:
Thanks for the commit

falkTX:
Please note that this breaks plugin host state whenever we decide to add some parameters.
The plugin DSP side is not allowed to change its own input parameters.
For programs, DPF provides an API to change them, which is exposed to the host format as appropriate.

PR #9, opened 2017-07-17
first draft of macro documentation

ssj71:
just getting the conversation started. More to be added as I actually play with the thing (and some screenshots).


ssj71:
updated this. Its as good as I have time for in 3.0.2. You may need/want to compress the images and/or switch the directory they're in. Hope this helps. I've gotta wrap up the announcement now.

PR #10, opened 2017-10-14
Add Polish l10n for .desktop files

m4sk1n:


fundamental:
Thanks for the commit

PR #11, opened 2018-02-10
Modulation volume representation from int to float

dsheeler:


fundamental:
Changes are needed to alter the saved data. You'll want to use xml.addparreal when saving. Once this is done, then you'll need to update the test data file which should be something like src/Tests/guitar???.xmz

fundamental:
PluginTest runs the tests for a full application save and load among several other things, so failures there will occur when full application XML serializations change.

dsheeler:
How do I update the test data file?  I tried:

% ./src/zynaddsubfx --load=../src/Tests/guitar-adnote.xmz --dump-oscdoc=guitar-new.xmz

But replacing guitar-adnote.xmz with new file produced lots more failures.

fundamental:
Typically for these sorts of small changes I look at the output of PluginTest which will display a diff of what's new. Then I manually update the test file. You can recreate the test file by loading it and then saving it. Note that --dump-oscdoc does not output a save file, but it outputs metadata about Zyn parameters which is used to generate http://zynaddsubfx.sourceforge.net/zyn-ports/ .

If you load the XMZ and save another  XMZ (master settings file) then you'll likely have a new file with some level of gzip compression (the compression level is a setting which is currently not exposed in the zyn-fusion GUI). To get plain text you can use gunzip. That's more complex than looking at the test output and updating by hand IMO though.

PR #12, opened 2018-03-15
Move the generation of chunked html pages to asciidoctor

aerostitch:
As mentionned in https://sourceforge.net/p/zynaddsubfx/code/merge-requests/6/


fundamental:
Thanks for the upgrade

PR #14, opened 2018-05-25
desktop launcher for multi-channel jack outputs

nick87720z:
Would be handy to initially start zyn with per-instrument channels.


fundamental:
Thanks for the addition

PR #15, opened 2018-05-28
fix path for save-osc test

nick87720z:
Current version doesn't work if buildir not in root of source dir


fundamental:
I had to restart the travis-CI builder (likely sourceforge instability for fetching a copy of liblo), but it seems like it runs cleanly now.

PR #16, opened 2018-07-14
Save and load slot names

pdesaulniers:
Should fix https://github.com/zynaddsubfx/zyn-fusion-issues/issues/173. I might not have done this correctly :p


fundamental:
It looks about right.
Have you tested it?

pdesaulniers:
Yes, it does seem to work correctly.

fundamental:
Sounds good to me then.

PR #19, opened 2018-12-11
LFOParams: Add 'RANDOM' LFO shape in the rtosc::Ports structure

pdesaulniers:
Adds the RAN LFO shape to the generated schema.


pdesaulniers:
Mmh, I've just realised that there is enough space in the UI to call it 'Random' instead. That would be better.

pdesaulniers:
Should be good now.

PR #20, opened 2018-12-19
Tweak filter defaults

pdesaulniers:
Should fix https://github.com/zynaddsubfx/zyn-fusion-issues/issues/99 .


fundamental:
Looking good. I'll double check the PR later this evening to verify there's nothing I've missed.

fundamental:
I haven't though of any negative consequences, so merged.

PR #23, opened 2019-03-24
convert subpar/ampvelocityscale to float

michiboo:
Hi I convert amp velocity scale function to float in subpar!


fundamental:
It looks like this PR is missing the ability to load older versions of the ampvelocityscale parameter to maintain compatibility with older .xmz files.

michiboo:
Hi changes are made as requested!

fundamental:
Based upon the newer commit, it looks like the test file (guitar-adnote.xmz) was incorrectly updated. If it was previously at 90 (0..127), then it should have mapped to ~70.86 instead of staying at 90.

michiboo:
I am little bit confused as you told me before in test file it can contain different value other than default? Anyhow I will change it to 70.8 in testfile!

michiboo:
Hi, I have made changes as requested!

fundamental:
Everything is looking good to me. @dsheeler you don't see anything that I'm overlooking, right?

dsheeler:
@fundamental: looks good to me, too.

PR #27, opened 2019-04-12
Fix crash with mxml V3

schnitzeltony:
Loading my old qtractor-sessions started to crash on my Raspi. A remote debug
session lead to zynaddsubfx and the point of code:

qtractor-crash

Wih this modification the sessions load again without issues.


fundamental:
Thanks for the fix and providing the background as to the source of the error.
It looks like there's likely a bug upstream in MXML if it's returning NULL for an attribute, but the patch seems to be the best fix possible on Zyn's side. :+1:

PR #28, opened 2019-04-20
change padpar parameter

dhruveshk:


fundamental:
Current commits do not compile due to an incorrect macro name (I think it's rDoc, not rdoc). The tests cannot be run, but I do expect at least one test to fail once the compilation bug is fixed (PluginTest).
The version check would not be against 3.0.3 any more since there have been subsequent releases.

fundamental:
Additionally, it would be great to incorporate the 4* into the dB mapping, adding an offset of +12.0411 dB into the current -60..0 dB range.

dhruveshk:
> Additionally, it would be great to incorporate the 4* into the dB mapping, adding an offset of +12.0411 dB into the current -60..0 dB range.
ok, will be doing it

fundamental:
Bump

PR #29, opened 2019-05-03
Fix BashCompletion.cmake output script

nick87720z:
BASHCOMP_PKG_PATH is supposed to replace BASHCOMP_USER_PATH if set, rather than be existing file/dir, since it is about to be created.


fundamental:
As per the IRC discussion I applied the slightly older patch from the sourceforge tracker. Let me know if this one is still needed.

nick87720z:
This PR is about different feature - to override BASHCOMP_PKG_PATH. It changes different line, just before block, fixed by that another patch.

fundamental:
Merged

PR #30, opened 2019-06-07
(WIP) Adding Watch Points for signal scope

michiboo:
To Do list:

- [x] Add watch point as proof of concept  

- [ ]  Testing

- [x] Add watch point to ADnote

- [x] Add watch point to SUBnote

- [x] Add watch point to PADnote

- [ ] Find the best way to add large number of watch points

### DIFF
--- a/src/Misc/Master.cpp
+++ b/src/Misc/Master.cpp
@@ -359,7 +359,7 @@ static const Ports automate_ports = {
 static const Ports watchPorts = {
     {"add:s", rDoc("Add synthesis state to watch"), 0,
         rBegin;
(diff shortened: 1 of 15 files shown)


fundamental:
Based upon reading the commits, it does not look like multiple notes or multi-frame sampling is currently considered. See attached timing diagram for some elaboration.

2019-06-watchpoint-timing

fundamental:
I tidied up the unit test somewhat and the failing steps in the unit test should show you some of the current problems. Once those are resolved feel free to ping me and I can generate the exact samples output to the test to verify that there's no difference between our systems.

PR #32, opened 2019-07-23
fix trigger test by changing position of watchpoint in SUBnote

michiboo:


fundamental:
Here's a patch to fix inaccurate capturing of the previous sample which seems to correct some triggering reliability issues.
[accurate-prev-sample.diff](https://github.com/zynaddsubfx/zynaddsubfx/files/3439160/accurate-prev-sample.txt)

fundamental:
Merged via a squashed commit since multiple 'fix X' and 'patch X' commits would have left the commit history more confusing than need be.

PR #33, opened 2019-07-28
Dead backend fixes

JohannesLorenz:
Can you please review?

When done, please merge with a **regular merge commit**, rather than squashing or rebasing.

### DIFF
--- a/src/Misc/Master.cpp
+++ b/src/Misc/Master.cpp
@@ -1123,28 +1123,43 @@ int msg_id=0;
 
 bool Master::runOSC(float *outl, float *outr, bool offline)
 {
-    //Handle user events
-    char loc_buf[1024];
-    DataObj d{loc_buf, 1024, this, bToU};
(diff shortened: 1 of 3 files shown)


fundamental:
"MiddleWare remains with a dangling master pointer" isn't right
according to my memory.

From what I recall would happen is in offline mode you would have

0 - master A
1 - master A
1.5 - load new master B
1.75 - set middleware's new master to B and send the update message
       assuming that master A is going to receive the update
1.85 - master B
2 - master B receives an update to itself
3 - master B frees itself and tries to update the system to master B
4 - master B (now invalid memory while master A is leaked)


I think the code of the second commit is sound, though the commit
message is a bit misleading in that one statement.
Would you agree with my reading or am I mistaken?

JohannesLorenz:
I just force-pushed the same commit with clearer commit message, naming the exact member `MiddleWareImpl::master`.

Feel free to push the button whenever you have time, please just do a real merge commit and don't delete the branch.

PR #34, opened 2019-08-08
Moderinizations for READMEs and NEWS file, http link upgraded to https where possible

Jacalz:
Here's a lot of changes to the general documentation:
- Converted HISTORY.txt to markdown and renamed to NEWS.md.
- Removed the old Changelog file (all releases are in NEWS.md).
- Add hyperlinks to all valid git tags from NEWS file.
- Rewrote all README files in markdown and also the TODO-release file.
- A lot of links (almost all) have been upgraded to https where possible.
- Minor link updates and wording updates trough out documentation.

On a further note, are the doc files really used in the repo? Seems like they are very outdated and that the new repo [user-manual](https://github.com/zynaddsubfx/user-manual) has the up to date version. Should I remove the whole doc folder?


fundamental:
ZynAddSubFX uses Asciidoc (mostly through asciidoctor at this point). As such no markdown for Zyn specific resources should be used.

fundamental:
In general I'm happy to see links updated, but otherwise I don't see the other changes as particularly beneficial. The general documentation is moving to the user-manual, however the move has not been completed and as such a good portion of the documentation is waiting to be migrated from older sources.

Jacalz:
Not even the News file and main Readme? At the very least the news file is miles more readable through github. I understand if the readme files in the doc folder need asciidoc files through.

My proposal is to have the main readme and the news file in markdown and I will leave everything else’s in asciidoc if that’s okay?

On a further note, are there any specific parts of the documentation that i could help move over? :)

Jacalz:
I mean if the docs are already in a half deprecated state, having the main readme and the news file in markdown would hurt much right?

fundamental:
The news file is also available at http://zynaddsubfx.sourceforge.net/news.html where I currently assume most users would end up reading it. If the in-repo on is converted to non-plain text in the form of Asciidoc (which github has a renderer for), then I would suggest changing the links for each tag as the current approach would likely mislead users into thinking that the downloads offered by github are valid. Unless github has fixed something relatively recently they are non-functional due to github's failure to include git submodules.

As per particular parts of the docs which need porting:

- it looks like there are some notes about which filter type to use which the new documents should discuss (rather than the current listing of what knob X does)
- I don't think the current midi information is copied over which would likely end up in an appendix
- I'm not sure that the LFOs have been tutorialized which the old docs have the reference information needed to start off with
- Effects are not discussed with motivating examples in the new docs
- System architecture is explicitly/intentionally not discussed in the manner that is discussed in parts of the old documents
- There's a section on non-linear functions from the older docs which could be reworked for the new docs
- More information from the even older docs may be relevant at some stage:
  - http://zynaddsubfx.sourceforge.net/doc_3.html
  - http://zynaddsubfx.sourceforge.net/doc_2.html
  - http://zynaddsubfx.sourceforge.net/doc_1.html
  - http://zynaddsubfx.sourceforge.net/doc_0.html
  - http://zynaddsubfx.sourceforge.net/doc/

Additionally a number of screenshots in the new documents need to be fully scripted with the UI's automated screenshot capture system. I've worked through a good few of them, but it would be ideal to capture as many as possible to basically make it impossible for them to drift out-of-date with the new UI as it's continued to be refined

Jacalz:
This should now (hopefully) be in a state where you can merge it. I have corrected what I believe needs to be corrected and done some further enhancements to the old history file (now NEWS.txt) and to the old README file. I reverted back the zyn specific markdown pages to .txt and rewrote the worting in the authors file.

fundamental:
Excellent. Thanks for taking the time to help update our docs. :+1:

Jacalz:
You are welcome! I’m happy to be helping out and I might come back with further improvements in due time 🙂

PR #36, opened 2019-08-12
(WIP) Extend watch points

michiboo:


fundamental:
If possible when adding additional watchpoints, naming the class member variables to something more distinctive might help verify that the correct watch point is collecting data at the right point. The current trend will result in there being watchpoint1,watchpoint2, ... watchpointN which doesn't say in the variable name what they're doing relative to eachother.

fundamental:
Not a blocker for merging, but it looks like the add synth pre and post mix signals are showing up as the same. I tried messing around with unison which seemed to be handled between the two points.

What's your setup for seeing differences between those two points?

PR #38, opened 2019-09-15
Bugfix: Removes occurences to HISTORY.txt

hickscorp:
The build currently fails because of a missing file at the root of the build tree. This PR fixes it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,7 +35,7 @@ include(CTestConfig.cmake)
 add_subdirectory(src)
 add_subdirectory(doc) # Doxygen only
 
-install(FILES AUTHORS.txt COPYING HISTORY.txt README.adoc
+install(FILES AUTHORS.txt COPYING README.adoc
 	DESTINATION share/doc/zynaddsubfx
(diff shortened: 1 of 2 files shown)


jofemodo:
I've tested and the problem really exists (make install fails!!) and it gets solved with this PR.

Jacalz:
I'm sorry this is my fault. I didn't know that the file was mentioned in that file.

fundamental:
Thanks for the fix :+1:

hselasky:
Hi I see that the sourceforge tarball has a HISTORY.txt file should this be installed?

Looks like someone has to resolve this merge conflict:

https://sourceforge.net/p/zynaddsubfx/code/ci/98e0b77a1a543252307dffb4a32abb2f852c8618

--HPS

fundamental:
On 10-03, HPS wrote:
> Hi I see that the sourceforge tarball has a HISTORY.txt file should this be installed?

Previous versions used the HISTORY.txt file, though it seems that going
forward one would not be used.
I've resolved the merge conflict running under that assumption.

PR #39, opened 2019-10-17
Add microtonal support

piegamesde:
Is there a way to overload functions in the OSC dispatcher? I first tried `noteOn:iiif`, but If I'd send an event, it would also trigger `noteOn:iii`, which is quite annoying.

### DIFF
--- a/src/Misc/Master.cpp
+++ b/src/Misc/Master.cpp
@@ -477,10 +477,16 @@ static const Ports master_ports = {
             sysefsendto.dispatch(msg, d);
         }},
 
-    {"noteOn:iii", rDoc("Noteon Event"), 0,
+    {"noteOn:iii:iiif", rDoc("Noteon Event"), 0,
         [](const char *m,RtData &d){
             Master *M =  (Master*)d.obj;
-            M->noteOn(rtosc_argument(m,0).i,rtosc_argument(m,1).i,rtosc_argument(m,2).i);}},
+            if (rtosc_narguments(m) > 3)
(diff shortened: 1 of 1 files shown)


fundamental:
The dispatcher should work with ports X:i and X:ii, though there may have been a change in that when the perfect minimal hashing was added as a speed improvement. Typically ports are made with multiple types that they accept e.g. X::i:ii would evaluate to a path with name X and arguments of "" (nothing), "i" (one int), "ii" (two ints). Within the handler/callback-function the behavior would change based upon the rtosc_argument_string() which would return what the message actually contained.

fundamental:
@JohannesLorenz do you know why the port checker is failing here? It doesn't look like this change should impact it.

JohannesLorenz:
@fundamental It's caused by the rtosc submodule update, even with the previous zyn version and the current rtosc it fails. This is my fault. Investigating now...

fundamental:
@piegamesde  In that case can you modify the commit to only include the change to Master.cpp? That way the continuous integration build passes? In that case it should be ready to merge.

piegamesde:
I hope you don't mind me force-pushing over and over again :)

JohannesLorenz:
This zyn bug has already been fixed in c40d5014c75be85e562392ee0565e723c490a3cd. If you rebase to a current master (which you somehow did now), all CI errors will be gone.

fundamental:
Thanks for the patch and for bearing with the delays.

PR #40, opened 2019-10-23
Fix abort on pgm change to undefined part

mindhog:
This fixes a bug where zyn aborts after a midi program change to an undefined patch.  I can consistently reproduce this by sending midi bank change MSB, bank change LSB and then a program change to a program that is not defined in the bank.  For some reason this does not occur when sending a just a program change for a program not defined in the current bank on startup.

---
This change is [Reviewable](https://reviewable.io/reviews/zynaddsubfx/zynaddsubfx/40)

### DIFF
--- a/src/Misc/MiddleWare.cpp
+++ b/src/Misc/MiddleWare.cpp
@@ -471,12 +471,7 @@ public:
         if(actual_load[npart] != pending_load[npart])
             return;
         assert(actual_load[npart] <= pending_load[npart]);
-
-        if(!filename || !*filename){
-            fprintf(stderr, "Warning: failed to load undefined part<%s>\n",
-                    filename);
-            return;
-        }
+        assert(filename);
 
         //load part in async fashion when possible
 #ifndef WIN32
@@ -1582,6 +1577,11 @@ static rtosc::Ports middlewareReplyPorts = {
         Bank &bank        = impl.master->bank;
         const int part    = rtosc_argument(msg, 0).i;
         const int program = rtosc_argument(msg, 1).i + 128*bank.bank_lsb;
+        if (program >= BANK_SIZE) {
+            fprintf(stderr, "bank:program number %d:%d is out of range.",
+                    program >> 7, program & 0x7f);
+            return;
+        }
         const char *fn  = impl.master->bank.ins[program].filename.c_str();
         impl.loadPart(part, fn, impl.master, d);
         impl.uToB->write(("/part"+to_s(part)+"/Pname").c_str(), "s",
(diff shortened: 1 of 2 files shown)


fundamental:
Thanks for the bug fix. As these sorts of bugs can be hard to spot if a regression occurs, we like to use unit tests to verify that the incorrect behavior doesn't occur.

Could you add a test which shows the error in https://github.com/zynaddsubfx/zynaddsubfx/blob/master/src/Tests/MiddlewareTest.h or https://github.com/zynaddsubfx/zynaddsubfx/blob/master/src/Tests/MessageTest.h ? The former would be for simpler communications which don't need the realtime thread to run and the latter for ones where both threads need to be running to replicate the issue.

Based upon the code fix it does look like there might be some related bugs which the unit test could help highlight.

mindhog:
I was hoping you'd ask for that :-)  I'll do so, but it may be a couple of days before I get back to this.

mindhog:
Can you advise me or point me some documentation on how to run MiddlewareTest?  "make" and "make test" don't seem to be doing it (everything builds and passes even if I add a syntax error to the file).

fundamental:
Sounds like you might not have CxxTest installed. If you're not sure try running mkdir build2 && cd build2 && cmake .. from within the zyn checkout. You should see a color coded listing of what libraries the cmake build system could find/enable.

mindhog:
So you're right, I didn't have cxxtest.  I installed it but still don't seem to be building the test (ran "cmake .; make clean; make; make test"  Here's what I'm currently missing:

-- Building on a 'Linux' System
-- PkgConfig -- found
-- zlib      -- found
-- mxml      -- found
-- fftw3     -- found
-- liblo     -- found
-- x11       -- found
-- xpm       -- NOT found
-- fltk      -- NOT found
-- ntk       -- found
-- x11       -- found
-- OSS       -- found
-- ALSA      -- found
-- JACK      -- found
-- PA        -- NOT found
-- Lash      -- NOT found
-- DSSI      -- NOT found
-- CxxTest   -- found
-- Lash      -- NOT enabled
-- DSSI      -- NOT enabled
-- tests     -- NOT enabled
-- ALSA      -- enabled
-- JACK      -- enabled
-- OSS       -- enabled
-- PA        -- NOT enabled
-- c++ async -- usable

I see tests are still not enabled.  Do I need to install all dependencies to build the tests?

mindhog:
Ok, I think I've got it.  I needed to clear out the cmake cache (sorry, I don't really have much experience with cmake).  All good now.

mindhog:
I've added a unit test and improved the fix.  PTAL?

fundamental:
Thanks for the inclusion of a unit test. While the comment does mention possible issues if the codebase changes in the future, using valigrind with the test suite should still be able to detect a regression. Thanks for taking the time to work through the process of adding a test :+1:

PR #43, opened 2019-12-23
fix in distorsion function 7 (limiter) and new waveshapers in oscilgen

friedolino78:
now with corrected rtosc dependency


JohannesLorenz:
Thanks!

PR #45, opened 2019-12-23
changed ENV time parameters to float

friedolino78:


JohannesLorenz:
@friedolino78 Is this PR ready to review, or do you want to add more commits?

friedolino78:
> @friedolino78 Is this PR ready to review, or do you want to add more commits?

it was discussed in: [https://sourceforge.net/p/zynaddsubfx/patches/40/](https://sourceforge.net/p/zynaddsubfx/patches/40/)

unit test needs to be fitted. i will have a look at it.

friedolino78:
ready for review, i guess :-)

src/Params/EnvelopeParams.cpp
```diff
@@ -114,8 +158,27 @@ static const rtosc::Ports localPorts = {
         d.reply(d.loc, "i", env->Envmode);
         rEnd},
 
-    {"envdt", rDoc("Envelope Delay Times"), NULL,
+    {"envdt", rDoc("Envelope Delay Times (log)"), NULL,
```
fundamental: ms, not log, right?

src/Params/EnvelopeParams.cpp
```diff
@@ -114,8 +158,27 @@ static const rtosc::Ports localPorts = {
...
+                arg_types[i] = 'f';
+            }
+            d.replyArray(d.loc, arg_types, args);
+        } else {
+            for(int i=0; i<N && i<M; ++i) {
+                env->envdt[i] = (rtosc_argument(msg, i).f)/1000; //store as seconds in member variable
+            }
+        }
+        rEnd},
+    {"dt", rDoc("Envelope Delay Times (lin)"), NULL,
```
fundamental: seconds not linear, right?

src/Params/EnvelopeParams.cpp
```diff
@@ -496,8 +555,15 @@ void EnvelopeParams::getfromXML(XMLwrapper& xml)
     for(int i = 0; i < Penvpoints; ++i) {
         if(xml.enterbranch("POINT", i) == 0)
             continue;
-        if(i != 0)
-            Penvdt[i] = xml.getpar127("dt", Penvdt[i]);
+        if(i != 0) {
+            if(!xml.hasparreal("A_dt")) {
```
fundamental: This seems to be the only actual bug. It should be xml.hasparreal("dt") unless I'm mistaken.

PR #46, opened 2020-01-19
Optional LP filter for RND and SQR LFO (127=off)

friedolino78:


fundamental:
While testing, I was able to get NaN values out of the LFO fairly easily by changing from Pcutoff=127 to Pcutoff != 127 while there was a running note (try printing out the 'output' in biquad() and you'll see that in some setups it goes to +-inf fairly easily).

In the second commit I see the guitar adnote file is duplicated under a different name.

In the first commit I see that the parameter is defined in the older 0..127 fashion. Is that intended for the final implementation?

Is the usage in RND/square only intended?

The filtering is defined without referencing the overall system sample rate or the buffersize and as such will behave differently when those configurations are changed. If this is merged in, the filter coefficients must behave the same (acoustically) if the user changes the buffer size or the sample rate.

It does produce some pretty fun to play with variations with varying the cutoff in the amplitude LFO. I was somewhat unsuccessful in using it in the frequency/filter LFOs, but part of that was trying to track down the crashes which were generated when the LFO started to output NaN values.

friedolino78:
I could not reproduce nans on the biquad() output. Can you tell me which preset you used and what  the LFO was applied to?

for the datatype the 127 steps are no problem as the parameter is not as sensitive as others. but if it has no performance drawback I will change it to float and a more physical intuitive range/unit.

fundamental:
I'm putting this back on my todo list. I should be able to create at least one unit test which exhibits the problem that I was able to observe.

fundamental:
```
/*
  ZynAddSubFX - a software synthesizer
  LfoTest.h - CxxTest for Lfo limits
  Copyright (C) 2020 Mark McCurry
  Authors: Mark McCurry
  This program is free software; you can redistribute it and/or
  modify it under the terms of the GNU General Public License
  as published by the Free Software Foundation; either version 2
  of the License, or (at your option) any later version.
*/


#include <cxxtest/TestSuite.h>
#include "../Synth/LFO.h"
#include "../Params/LFOParams.h"
#include "../globals.h"

using namespace std;
using namespace zyn;

SYNTH_T *synth;


class AdNoteTest:public CxxTest::TestSuite
{
    public:
        SYNTH_T      *synth;
        AbsTime      *time;
        LFO          *lfo;
        LFOParams    *lfop;
        int randval(int min, int max)
        {
            int ret = rand()%(1+max-min)+min;
            //printf("ret = %d (%d..%d)\n",ret, min,max);
            return ret;
        }

        void setUp() {
        }

        void tearDown() {
        }

        void randomize_params(void) {
            lfop->Pintensity  = randval(0,255);
            lfop->Pstartphase = randval(0,255);
            lfop->Pcutoff     = randval(0,255);
            lfop->PLFOtype    = randval(0,7);
            lfop->Prandomness = randval(0,255);
            lfop->Pfreqrand   = randval(0,255);
            lfop->Pcontinous  = randval(0,1);
            lfop->Pstretch    = randval(0,255);
        }

        void run_initialization(void)
        {
            lfo  = new LFO(*lfop, 440.0f, *time);
            for(int i=0; i<100; ++i) {
                float out = lfo->lfoout();
                if(!(-1.0f < out && out < 1.0f))
                    printf("%f\n", out);
            }
            randomize_params();
            for(int i=0; i<100; ++i) {
                float out = lfo->lfoout();
                if(!(-1.0f < out && out < 1.0f))
                    printf("%f\n", out);
            }
            randomize_params();
            for(int i=0; i<100; ++i) {
                float out = lfo->lfoout();
                if(!(-1.0f < out && out < 1.0f))
                    printf("%f\n", out);
            }
            randomize_params();
            for(int i=0; i<100; ++i) {
                float out = lfo->lfoout();
                if(!(-1.0f < out && out < 1.0f))
                    printf("%f\n", out);
            }
        }

        void testDefaults() {
            synth = new SYNTH_T();
            time  = new AbsTime(*synth);

            lfop = new LFOParams();
            lfop->fel  = zyn::consumer_location_type_t::amp;
            lfop->freq = 2.0f;
            lfop->delay = 0.0f;
            for(int i=0; i<10000; ++i) {
                randomize_params();
                run_initialization();
            }
        }
};
```

Modify this as needed, but this randomized testing should be sufficient to replicate the issue with the filter code.

friedolino78:
unfortunatelly it passes everytime. 
i replaced the adnote test code by yours, did make and make test. did i miss anything?

fundamental:
The test does not have any assert statements to fail at, but if you run the test directly (the resulting binary should be in build/src/Tests/ ), then you should see any output which exceed the expected -1.0..1.0 range.

friedolino78:
> The test does not have any assert statements to fail at, but if you run the test directly (the resulting binary should be in build/src/Tests/ ), then you should see any output which exceed the expected -1.0..1.0 range.

did id. no such output. i will try the assert.

fundamental:
@JohannesLorenz if you have a chance it would be good to know if I'm providing a test which doesn't demonstrate a problem. Checkout the branch and use the above test and hopefully the problem can be replicated.

JohannesLorenz:
> @JohannesLorenz if you have a chance it would be good to know if I'm providing a test which doesn't demonstrate a problem. Checkout the branch and use the above test and hopefully the problem can be replicated.

When I do this, I get more than 600,000 lines of output not in range [-1, 1].

friedolino78:
I took your test code, replaced the printf by TS_ASSERT and added it to existing AdNoteTest.
in future i could imagine splitting the adnote test and creating own tests for LFO and ENV.

friedolino78:
with this it is possible to create frequency jitter like on this video:
[https://www.presetpatch.com/article/vco-harmonic-jitter](https://www.presetpatch.com/article/vco-harmonic-jitter)

fundamental:
Finally looping back to this one. Looking at the above discussion things should be good, though I'll do a sanity check (hopefully Monday morning) and then merge. Let me know if I'm forgetting anything.

fundamental:
I'm seeing some unexpected behavior with LP around '111' or so (RND mode, high depth, moderate freq, amp LFO, buffersize=256). Not a NaN and to be fair it sounds good with that behavior, but I want to make sure I understand what's happening.

friedolino78:
> I'm seeing some unexpected behavior with LP around '111' or so (RND mode, high depth, moderate freq, amp LFO, buffersize=256). Not a NaN and to be fair it sounds good with that behavior, but I want to make sure I understand what's happening.

with cutoff frequencies around Fs/2 this biquad starts to oscillate, which can lead to audible artefacts when used in amp or freq lfo. with the last commit the param is mapped to a range up to 0.4
tests showed at setting 126 edges are still quite steep.

src/Synth/LFO.cpp
```diff
@@ -126,31 +126,31 @@ float LFO::baseOut(const char waveShape, const float phase)
...
-	    z1 = input * a1 + z2 - b1 * output;
-	    z2 = input * a2 - b2 * output;
-	}
-    return (cutoff==127) ? input : output;
+        cutoff = lfopars_.Pcutoff;
+        if (cutoff != 127) // at cutoff 127 we bypass filter, no coeffs needed
+        {
+            // calculate biquad coefficients
+            Fc = powf(cutoff + 7.0f, 2.0f)/ 45056.0f;
+            K = tan(PI * Fc);
```
fundamental, line 137: I'm confused. Previously dt_ was used as a mechanism to make sure that the cutoff frequency of the filter was independent of samplerate/buffer size, but removing any dependent on dt_ seems like it would make things dependent on what buffer size and sample rate the user selected.

friedolino78: now i am confused, too. don't know why i kicked it out.
i think you are right.

src/Synth/LFO.cpp
```diff
@@ -21,6 +21,8 @@
 
 namespace zyn {
 
+#define LIMIT(x,lu, ll) (x>lu ? lu : (x<ll ? ll : x) )
```
fundamental, line 24: LFO.cpp looks like it already includes Util.h. Util.h defines T limit<T>(T val, T min, T max). It seems like for consistency it should be used instead of another definition.

fundamental:
Looks good, seems like the frequency issue is resolved and the random parameter test ensures there should be no excessively loud sounds. The LIMIT vs limit comment is mostly style, but with that fixed, I'd say this is 100% merge-ready.

PR #47, opened 2020-02-18
Icon() expected Fl_RGB_Image and received Pixmap. FLTK icons work now.

thenorili:


thenorili:
Probably the cause of #https://github.com/zynaddsubfx/zyn-fusion-build/issues/30

src/UI/MasterUI.fl
```diff
@@ -87,6 +87,22 @@ decl {\#if !defined(PLUGINVERSION) && HAS_X11
...
 } 
 
+decl {\#if !defined(PLUGINVERSION) && HAS_X11
+\#include <FL/Fl_RGB_Image.H>
+\#endif} {private local
+} 
+
+decl {\#if !defined(PLUGINVERSION) && HAS_X11
+\#include <FL/FL.H>
+\#include "Enumerations.H"
```
fundamental: That should be #include <FL/Enumerations.H>, right?

thenorili: Enumerations.H is a file, not a class. In the header files and example code I read, it was included in this manner.

thenorili: I found a piece of example code (/test/sudoku.cxx) that includes it in this manner, so I'll make the change.

thenorili:
I'm going to review it further, but it may be true that the XCreatePixmapFromData call was the only function call to X11/xpm.H and it can be removed from the codebase.

fundamental:
Thanks for the update :+1:

PR #49, opened 2020-03-10
part stores its xiz filename, /savexml saves there

friedolino78:


JohannesLorenz:
I'll go for the review :smile:

src/Misc/Part.cpp
```diff
@@ -1038,6 +1042,7 @@ int Part::saveXML(const char *filename)
     xml.endbranch();
 
     int result = xml.saveXMLfile(filename, gzip_compression);
+    strcpy(loaded_file,filename);
```
JohannesLorenz, line 1045: Here, and also in `loadXMLinstrument`, I'd rather user `strncpy` for safety. If you change it, please make sure it's null terminated: `Warning: If there is no null byte among the first n bytes of src, the string placed in dest will not be null-terminated. (man page)`

src/Misc/Part.cpp
```diff
@@ -165,8 +165,12 @@ static const Ports partPorts = {
 
             //d.broadcast("/damage", "s", part_loc);
         }},
-
-
+        {"savexml:", rProp(internal) rDoc("Save Part to the file it has been loaded from"), 0,
+        [](const char *, RtData &d)
+        {
+            Part *p = (Part*)d.obj;
+            p->saveXML(p->loaded_file);
```
JohannesLorenz, line 172: If `p->loaded_file` has never been used, I get `Conditional jump or move depends on uninitialised value(s)` from valgrind.

fundamental:
Per the goal of the PR, it originally came up in IRC as a way of tracking what instrument was loaded so that it could be displayed in the bank view of the UI more easily.

friedolino78:
my primary goal was to make the save button in zest work in a way like in most applications.
save to the file you loaded from or you saved to before.
but what has to be done, if the name was changed in the part settings? I imagine, that in that case a new named file with the same number prefix should replace the old one.

src/Misc/Part.cpp
friedolino78: > If `p->loaded_file` has never been used, I get `Conditional jump or move depends on uninitialised value(s)` from valgrind.

yes, it doesn't work with a new preset, that has never been saved. easiest way would be not saving at all in such a case. 

to prevent loss of creative work it could be saved to the first free slot in a bank reserved for this case.
slightly modified this could serve another use case i have thought of for a long time: an easy way to save the steps of your creative process.
In the old gui you could do this by using the write button on empty slots.

src/Misc/Part.cpp
friedolino78: question is where does the n come from. 
in case of loading the filename comes from osc:

Middleware.spp:1432
const char *file  = rtosc_argument(msg,1).s;

JohannesLorenz: The size of the source string is not important, the size of the destination string (`sizeof(loaded_file)`) is, since we don't want to overwrite the destination string.

src/Misc/Part.cpp
JohannesLorenz: @fundamental What's your opinion about this suggestion?

src/Misc/Part.cpp
friedolino78: ok

src/Misc/Part.cpp
friedolino78: New idea: if the part has never been loaded or saved, the savexml port handler makes up a new one from date and time.
still open question: what if i press save again after editing. new filename or overwrite the old file. I tend to overwrite only files that i have loaded or given a name myself.

fundamental: With regards to the suggestion, there's certainly some flexibility in the new design as not a lot of work has been put into making it friendly to save instruments to banks. Having one slot reserved as a defacto work in progress slot is interesting. Not sure if users would like it or not, but I'm not raising any immediate red flags. IMO the major thing to watch out for is creating a workflow which results in users unintentionally overwriting an old patch while expecting to save it elsewhere. Saving based upon date/time combinations is one approach.

For many users you have the difference of a bank that they control and a bank which they've received i.e. during normal operation a bank is either read only or it is editable by a user. Perhaps some way of expressing that would be good, though I'm unsure if this is too big of an issue to discuss given the current scope of this PR.

friedolino78: Indeed I didnt think of the built-in-preset use case...

I think the read only could be accomplished by filesystem permissions.
-> if a read-only file is loaded, the save button should be disabled.

option 2: in linux the built-in vs user banks can be separated by using different directories /usr/ vs /home/. is there something similar on windows?
in built-in banks the save button should also be disabled.

fundamental:
@JohannesLorenz Since you started the review I'll let you give the final OK, if that's alright with you.

friedolino78:
I think, the only point, thats still open is how to protect the built in preset banks. but that's the same with saving via write mode. 
other than that the current commits doesnt seem to break anything because the date based filenames beginning with 2020 are beyond the slots of the banks.

src/Misc/Part.h
```diff
@@ -159,6 +159,8 @@ class Part
         bool Pefxbypass[NUM_PART_EFX]; //if the effects are bypassed
 
         int lastnote;
+        char loaded_file[256];
+        bool hasFilename;
```
JohannesLorenz, line 163: Is this variable necessary? Could you not just initialize `loaded_file` as an empty string and then check if `(*loaded_file == 0)`?

src/Misc/Part.cpp
```diff
@@ -1038,6 +1050,8 @@ int Part::saveXML(const char *filename)
     xml.endbranch();
 
     int result = xml.saveXMLfile(filename, gzip_compression);
+    strncpy(loaded_file,filename, sizeof(loaded_file));
```
JohannesLorenz, line 1053: Like I mentioned in the now outdated comment, if the length of `filename` will be equally long to/longer than `sizeof(loaded_file)`, this will result in a non-null-terminated string and can lead to crashes. See the `strncpy` man page.

src/Misc/Part.cpp
```diff
@@ -165,8 +165,19 @@ static const Ports partPorts = {
...
-
+        {"savexml:", rProp(internal) rDoc("Save Part to the file it has been loaded from"), 0,
+        [](const char *, RtData &d)
+        {
+            Part *p = (Part*)d.obj;
+            if (!p->hasFilename) {  // if part was never loaded or saved
+                time_t rawtime;     // make a new name from date and time
+                struct tm * timeinfo;
+                time (&rawtime);
+                timeinfo = localtime (&rawtime);
```
JohannesLorenz, line 176: You could just write `const struct tm* timeinfo = localtime (&rawtime);` to save a line of code.

src/Misc/Part.cpp
JohannesLorenz: For me, load/save to/from disk and load/save to/from a bank are two different concepts. I never used banks in the past and I would not assume to find a not-saved file there. I would not like to learn the banks concept just because I used this functionality without prior loading of the part.

I'd rather do it like in most applications: Of you click "save as", of course, always a dialogue comes. If you click save and it has not yet been loaded/saved, again a dialogue, but if it has, direct saving. In this case, zyn send the GUI a boolean in return to say "I could not do a 'save'", and the UI can then raise a dialogue (or an error window).

friedolino78: I work with banks because I use them in rehearsal room headless via midi only. if I notice that something needs to be altered, I write a note, change the bank back at home at my desktop and rsync it to the headless box.
My workflow is: load one preset from my bank (in browser or via PGM Change message),  change a few parameters unti I like it again and then save it to the same slot.

fundamental:
@JohannesLorenz I'm working through the backlog of zyn emails. Am I correct in reading that changes are currently requested from your review?

src/Misc/Part.h
friedolino78: good idea.

src/Misc/Part.cpp
friedolino78: done

friedolino78:
@JohannesLorenz I solved all your requests.

JohannesLorenz:
> Am I correct in reading that changes are currently requested from your review?

@fundamental That's indeed correct . Though @friedolino78 has solved them all after your comment - It's my turn now to check his rework.

src/Misc/Part.cpp
```diff
@@ -165,8 +165,18 @@ static const Ports partPorts = {
...
-
+        {"savexml:", rProp(internal) rDoc("Save Part to the file it has been loaded from"), 0,
+        [](const char *, RtData &d)
+        {
+            Part *p = (Part*)d.obj;
+            if (p->loaded_file[0] == '\0') {  // if part was never loaded or saved
+                time_t rawtime;     // make a new name from date and time
+                time (&rawtime);
+                const struct tm* timeinfo = localtime (&rawtime);
+                strftime (p->loaded_file,23,"%F_%R.xiz",timeinfo);
```
JohannesLorenz, line 176: Do you think we should print the filename to console, so the user knows how it has been saved? Or do you think it is obvious?

I'm OK with both.

friedolino78:
indeed locally I left a print that I had for debug purposes. and I still
like to see it. So I think it is a good idea.

Am 26. Juni 2020 21:43:05 schrieb Johannes Lorenz <notifications@github.com>:

PR #51, opened 2020-03-17
Warnings padsynth oscilgen

JohannesLorenz:
@fundamental Can you please check if the warning options here make sense to you and the resulting code changes look justified?

P.S.: This is a draft, please don't merge yet.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -182,11 +182,49 @@ endif()
 option (BuildForAMD_X86_64 "Build for AMD x86_64 system" OFF)
 option (BuildForCore2_X86_64 "Build for Intel Core2 x86_64 system" OFF)
 option (BuildForDebug "Include gdb debugging support" OFF)
+option (ExtendedWarnings "Enable all useful warnings" OFF)
 option (IncludeWhatYouUse "Check for useless includes" OFF)
 mark_as_advanced(IncludeWhatYouUse)
 
 set(CMAKE_BUILD_TYPE "Release")
 
+set (BuildOptions_ExtendedWarnings "")
+if (ExtendedWarnings)
+    if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+        # The following generates a list of warning exceptions
+        # (after it enables all warnings).
+        # This is for developers who want to make sure they really
+        # see all possible warnings, with a few exceptions that don't
+        # make sense in this code base.
+        # You can remove those exceptions if they don't complain too
+        # often about the code base. If you want to add exceptions,
+        # be VERY CAREFUL, ask the project maintainers for
+        # their consent.
+        set (BuildOptions_ExtendedWarnings_List
+        "-Wall" "-Wextra" "-Weverything" # start with really everything
+        "-Wno-c++98-compat" "-Wno-c++98-compat-extra-semi" "-Wno-c++98-compat-pedantic" # C++11 is a zyn build requirement
+        "-Wno-cast-align" # not relevant for our case
+        "-Wno-double-promotion" # Used too often to fix
+        "-Wno-old-style-cast" # forces you to use static_cast<> etc everywhere! too many errors
+        "-Wno-padded" # this tells you to realign your structs to have better memory management
+        "-Wno-zero-as-null-pointer-constant" # wants you to use nullptr in the whole code base
+        "-Wno-implicit-int-conversion" # too often to fix
+        "-Wno-extra-semi" # one semicolon more than useful - no danger to disable this
+        "-Wno-vla-extension" "-Wno-vla" # rtosc currently heavily uses them, and they are supported by our CI
+        "-Wno-exit-time-destructors" "-Wno-global-constructors" # used too often in rtosc
+        "-Wno-gnu-zero-variadic-macro-arguments" # used too often in zyn
+        "-Wno-shadow" # used too often in zyn
+        "-Wno-cast-align" # used too often in zyn
+        "-Wno-missing-prototypes" # used too often in zyn
+        "-Wno-sign-conversion" # used too often in zyn
+	"-Wno-float-equal" # can be disturbing in some cases
+        )
+        STRING(REPLACE ";" " " BuildOptions_ExtendedWarnings "${BuildOptions_ExtendedWarnings_List}")
+        MESSAGE(STATUS "Using extended Warning options: ${BuildOptions_ExtendedWarnings}")
+    else()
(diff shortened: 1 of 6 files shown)


src/CMakeLists.txt
```diff
@@ -182,11 +182,49 @@ endif()
 option (BuildForAMD_X86_64 "Build for AMD x86_64 system" OFF)
 option (BuildForCore2_X86_64 "Build for Intel Core2 x86_64 system" OFF)
 option (BuildForDebug "Include gdb debugging support" OFF)
+option (ExtendedWarnings "Enable all useful warnings" OFF)
```
fundamental: Looks good

src/DSP/FFTwrapper.h
```diff
@@ -24,7 +24,7 @@ class FFTwrapper
 {
     public:
         /**Constructor
-         * @param fftsize The size of samples to be fed to fftw*/
+         * @param fftsize_ The size of samples to be fed to fftw*/
```
fundamental: Yep

src/DSP/FFTwrapper.h
```diff
@@ -50,10 +50,10 @@ template<class _Tp>
 std::complex<_Tp>
 FFTpolar(const _Tp& __rho, const _Tp& __theta = _Tp(0))
 {
-        _Tp __x = __rho * cos(__theta);
+        _Tp __x = __rho * std::cos(__theta);
```
fundamental: Something in the back of my mind indicates that there was a difference between cos and std::cos on some platform (e.g. window with MSVC). It might be worth a quick google, but otherwise it should be fine.

src/Misc/Time.h
```diff
@@ -43,7 +43,7 @@ class RelTime
             double deltaFrames = sec*t.framesPerSec();
             int64_t tmp = (int64_t)deltaFrames;
             frame = t.time() + tmp;
-            sample = t.samplesPerFrame()*(deltaFrames-tmp);
+            sample = (int32_t)(t.samplesPerFrame()*(deltaFrames-tmp));
```
fundamental: Sure

src/Params/FilterParams.h
```diff
@@ -52,13 +52,13 @@ class FilterParams:public PresetsArray
...
-        float    baseq;        //< Q parameters (resonance or bandwidth)
-        float    freqtracking; //< Tracking of center frequency with note frequency (percentage)
-        float    gain;         //< filter's output gain (dB)
+        unsigned Pcategory:2;  //!< Filter category (Analog/Formant/StVar)
+        unsigned Ptype:8;      //!< Filter type  (for analog lpf,hpf,bpf..)
+        unsigned Pstages:8;    //!< filter stages+1
+        float    basefreq;     //!< Base cutoff frequency (Hz)
+        float    baseq;        //!< Q parameters (resonance or bandwidth)
+        float    freqtracking; //!< Tracking of center frequency with note frequency (percentage)
+        float    gain;         //!< filter's output gain (dB)
```
fundamental: No problem here

src/Params/PADnoteParameters.cpp
```diff
@@ -898,12 +898,15 @@ int PADnoteParameters::sampleGenerator(PADnoteParameters::callback cb,
     float adj[samplemax];
     for(int nsample = 0; nsample < samplemax; ++nsample)
         adj[nsample] = (Pquality.oct + 1.0f) * (float)nsample / samplemax;
+    // QtCreator can't capture VLAs (QTCREATORBUG-23722), so this is
+    // a workaround to allow using the IDE
+    float * const adj_ptr = adj;
```
fundamental: Strange, but not a problem

src/Params/PADnoteParameters.cpp
```diff
@@ -948,10 +951,10 @@ int PADnoteParameters::sampleGenerator(PADnoteParameters::callback cb,
             float rms = 0.0f;
             for(int i = 0; i < samplesize; ++i)
                 rms += newsample.smp[i] * newsample.smp[i];
-            rms = sqrt(rms);
+            rms = sqrtf(rms);
             if(rms < 0.000001f)
                 rms = 1.0f;
-            rms *= sqrt(262144.0f / samplesize);//262144=2^18
+            rms *= sqrtf(262144.0f / samplesize);//262144=2^18
```
fundamental: All the func()->funcf() seems fine in this file.

src/Synth/OscilGen.cpp
```diff
@@ -1715,15 +1716,15 @@ FILTER(lp2)
 
 FILTER(hp2)
 {
-    if(par == 1)
+    if(par >= 1)
```
fundamental: Unless there's a good reason, 1.0 == 1.0, so there's no problem with a direct comparison of values.
As far as I can see the source of the 'par' value is going to produce exactly 1.0 in the case this is handling.

src/Synth/OscilGen.cpp
```diff
@@ -1327,9 +1328,9 @@ void OscilGen::add2XML(XMLwrapper& xml)
 
         xml.beginbranch("BASE_FUNCTION");
         for(int i = 1; i < synth.oscilsize / 2; ++i) {
-            float xc = basefuncFFTfreqs[i].real();
-            float xs = basefuncFFTfreqs[i].imag();
-            if((fabs(xs) > 1e-6f) || (fabs(xc) > 1e-6f)) {
+            float xc = (float)basefuncFFTfreqs[i].real();
+            float xs = (float)basefuncFFTfreqs[i].imag();
+            if((fabsf(xs) > 1e-6f) || (fabsf(xc) > 1e-6f)) {
```
fundamental: All float casts and func()->funcf() changes not otherwise noted in this file look correct.

src/Synth/OscilGen.cpp
JohannesLorenz: OK, but the compiler (which cannot know the background) complained that comparing floats with == was unsafe. We can either use a trick or just [disable `-Wfloat-equal`](https://stackoverflow.com/a/11422298), what do you think?

src/DSP/FFTwrapper.h
JohannesLorenz: I really couldn't find anything, neither for Win/MacOS, nor in general. I'll resolve this, feel free to re-open.

src/Synth/OscilGen.cpp
JohannesLorenz: Resolved by disabling `-Wfloat-equal`

PR #52, opened 2020-03-17
unison-size  minimum: 0 corrected to 1

strongheart1:
So, test.json in osc-bridge needs to be updated too, with the output from zynaddsubfx  -D option?
Like here in the version i just generated: [https://github.com/strongheart1/osc-bridge/commit/0d47f471b23d8a6106f1fa35a285641f435f1ebd](https://github.com/strongheart1/osc-bridge/commit/0d47f471b23d8a6106f1fa35a285641f435f1ebd) there are also other changes, due to envelope time parameter changes.


fundamental:
Thanks for tracking down the off-by-one error and correcting it. :+1:

PR #53, opened 2020-03-29
Fix PAD filter changes needing prepare

JohannesLorenz:
Before this, any realtime parameter change of PAD synth, e.g. filter,
triggered the "needPrepare" message, marking the PAD synth's "Apply
Changes" button.

### DIFF
--- a/src/Misc/MiddleWare.cpp
+++ b/src/Misc/MiddleWare.cpp
@@ -322,7 +322,7 @@ struct NonRtObjStore
                 strcpy(d.loc, obj_rl.c_str());
                 d.obj = pad;
                 PADnoteParameters::non_realtime_ports.dispatch(msg, d);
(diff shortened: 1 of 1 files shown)


JohannesLorenz:
> I think that the paste logic separately handles the preparation, however it may be something worth sanity checking.

OK, I'll just wait for your tests.

fundamental:
Reading through PresetExtractor.cpp, PADnoteParameters.cpp, and Oscilgen.cpp, it looks like the prepare flag should be handled correctly when the new xml is pasted into an oscilgen instance, so I think this should be good for a rebase and merge. Let me know if there was anything else which you wanted checked.

JohannesLorenz:
Nothing else to check imo. I'll rebase and merge. Thanks for the review!

PR #55, opened 2020-04-10
finish addsynth ENVs sustaining at 0 with 0 release to free up cpu

friedolino78:
This modification can free up some cpu time when playing with many sustained notes with many voices with different decay times. It should not have any sonical impact because it removes inaudible voices only.

### DIFF
--- a/src/Misc/Part.cpp
+++ b/src/Misc/Part.cpp
@@ -597,7 +597,7 @@ void Part::NoteOff(note_t note) //release the key
 }
 
 void Part::PolyphonicAftertouch(note_t note,
-				unsigned char velocity)
+                unsigned char velocity)
 {
     if(!Pnoteon || !inRange(note, Pminkey, Pmaxkey) || Pdrummode)
         return;
@@ -728,7 +728,7 @@ void Part::SetController(unsigned int type, note_t note, float value,
         break;
     case C_pitch: {
         if (getNoteLog2Freq(masterkeyshift, value) == false)
-	    break;
+        break;
 
         /* Make sure MonoMem's frequency information is kept up to date */
         if(!Ppolymode)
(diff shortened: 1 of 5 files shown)


fundamental:
Thanks for the addition, it should be a good compute saver. Just to head off any users with 'unique' free mode envelopes you could have something akin to (hopefully this asciiart renders correctly)

```
--+--|------------
-+-+-|-----*++---
+---+*+++*----*++
```

So as the user can define a custom envelope, they can have the sustain point at a volume level of zero, then increase the envelope post sustain, then eventually decrease it as expected.
It seems like the functionality targets A) volume envelopes only and B) is well behaved for non-custom envelopes. If you can add a quick check for A/B I can't think of any problems which would arise from the performance optimization.

src/Params/EnvelopeParams.h
```diff
@@ -19,6 +19,14 @@
 #include "Presets.h"
 
 namespace zyn {
+    
+   enum envmode_enum {
+        ADSR_lin,
```
fundamental, line 24: Looks like an off by one.
Since these were previously integer values saved to disk it seems reasonable to define the values explicitly. Perhaps I'm mistaken, but it looks like the values of the new enum do not match the old values.

src/UI/EnvelopeUI.fl
```diff
@@ -181,7 +181,7 @@ envfree->redraw();}
       }
       Fl_Button {} {
         label {Cancel}
-	tooltip {Cancel freemode editing}
+    tooltip {Cancel freemode editing}
```
fundamental, line 184: Looks like an unintentional whitespace change

src/Misc/Part.cpp
```diff
@@ -597,7 +597,7 @@ void Part::NoteOff(note_t note) //release the key
 }
 
 void Part::PolyphonicAftertouch(note_t note,
-				unsigned char velocity)
+                unsigned char velocity)
```
fundamental, line 600: Looks like an unintentional whitespace change (also same for the other change in Part.cpp within this commit).

fundamental:
Thanks for the improvement :+1:

PR #56, opened 2020-04-18
make /distorsion/waveform reflect the Pvolume par

friedolino78:
with this patch the waveform view in the distorsion effect reflects the effect of the AMT knob in zest (Pvolume par) in the way you hear it. At Pvolume=0 you hear no distorsion -> Transferfunction is linear. Now in waveview, too.


fundamental:
Looks good to me.

PR #57, opened 2020-04-21
performance improvements in distorsion and adnote

friedolino78:
lp/hp filter in distorsion effect are only calculated if cutoff freq parameter is not at max/min value.
slow powf operations in waveshaper replaced by multiplications
some precalculations in adsynth to speed up linerar interp oscillator

### DIFF
--- a/src/Synth/ADnote.cpp
+++ b/src/Synth/ADnote.cpp
@@ -1222,21 +1222,21 @@ inline void ADnote::ComputeVoiceOscillator_LinearInterpolation(int nvoice)
     Voice& vce = NoteVoicePar[nvoice];
     for(int k = 0; k < vce.unison_size; ++k) {
         int    poshi  = vce.oscposhi[k];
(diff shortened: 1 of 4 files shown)


src/Synth/ADnote.cpp
```diff
@@ -1213,21 +1213,21 @@ inline void ADnote::ComputeVoiceOscillator_LinearInterpolation(int nvoice)
 {
     for(int k = 0; k < unison_size[nvoice]; ++k) {
         int    poshi  = oscposhi[nvoice][k];
-        int    poslo  = (int)(oscposlo[nvoice][k] * (1<<24));
```
fundamental, line 1216: IMO the changes in this file should have some associated comments.
It's easier to understand the 1<<24, though it's understandable if the modifications make things faster.

friedolino78: I added some comments and changed the precalc value to float to prevent the implicit casting.

PR #58, opened 2020-04-23
parameter for square and cubic waveshaper

friedolino78:


friedolino78:
complement to https://github.com/mruby-zest/mruby-zest/pull/38

src/Misc/WaveShapeSmps.cpp
```diff
@@ -13,6 +13,7 @@
 
 #include "WaveShapeSmps.h"
 #include <cmath>
+#define X smps[i]
```
fundamental, line 16: Yeah, I'm going to recommend against this. If you have loops where the expression is used a number of times feel free to have "float &X = smps[i];", but listing this as a macro across the whole file makes it easy to mistakenly use and makes some of the code marginally less obvious.

That said, I do like that it makes some the further downwind changes easier to read, so as long as there's the "float &X = smps[i]" at the right scope, the smps[i]->X changes seem to be a net gain in readability.

fundamental:
After the macro is changed to a set of properly scoped references, this should be good to merge.

friedolino78:
the patches fe1754f and ff47100 has gone obsolete due to a less cpu intensive solution in [https://github.com/zynaddsubfx/zynaddsubfx/pull/62](https://github.com/zynaddsubfx/zynaddsubfx/pull/62)

friedolino78:
i guess for the float &X = smps[i] thing it's easier to make it new from current master.

fundamental:
I recall that other PR, but did that also add the different defaults as well? If so, then I guess this should be closed. If not then I guess it needs to be reworked?

friedolino78:
the different defaults are only needed for the proposed linearization parameters for square and cubic. In PR62 I solved the linearization more elegant and less cpu costly by simply allowing drive values < 1.0
So no parameters -> no default values.

fundamental:
Ah, ok. Just to make sure I'm not misinterpreting anything. Everything here is out-of-date and can be safely closed, right?

friedolino78:
yes. only missing thing would be reminding me to "&X = smps[i]" ;-)

PR #59, opened 2020-04-25
in fractional tunings, allow numerator/denominator greater than 127

jmckernon:
Tunings with numerators/denominators work during a session, but when loaded from a file, the values are limited to 0 <= x <= 127. This commit increases that limit to a safe one.

N.B. It's quite easy to end up with very complex fractions when tinkering with exotic tunings; see for instance the table at https://en.wikipedia.org/wiki/The_Well-Tuned_Piano#Tuning .


fundamental:
Looks like this should load correctly, saving is done with something compatible, and I can't immediately think of any negative consequence of larger ratio components.

Thanks for the improvement :+1:

PR #60, opened 2020-05-27
nprn controllers for part insert effects

friedolino78:
implementation of my proposal to use only midi channel 0 for master insertion effects and the other channels for the respective part's insertion effects.


fundamental:
partefx[] access can be out of bounds with the current commit as it has a different max value than the previous parlo value is constrained to. Resolve that and add a bit of a description to the commit message (in the general https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html style) and it should be good to go. I tend to read back through commit messages to ensure that change logs include any user visible features or bug fixes and this change is something that users may want to know about.

fundamental:

PR #62, opened 2020-06-07
reduce minimum drive of square/cubic distorsion

friedolino78:
drive values less than one (below CC value 44) together with offset make it possible to go though the transfer function at different spots. that allows in the result more gentle asymmetric transfer functions.


fundamental:
I guess it's good that these distortion types are still fairly new. That should mean that no patches of a fully released zyn version are impacted.

friedolino78:
yes. I actually think I am the only one using them ;-)
but i know what you mean.

PR #63, opened 2020-06-30
Fix error in ZLIB detection

carlo-bramini:
I'm getting this message on screen when CMake configures the project:

```
  The package name passed to `find_package_handle_standard_args` ;(ZLIB) does
  not match the name of the calling package ;(zlib).  This can lead to
  problems in calling ;code that expects `find_package` result variables
  (e.g., `_FOUND`) ;to follow a certain pattern.
Call Stack (most recent call first):
  /home/carlo/inst_cmake/share/cmake-3.17.3/Modules/FindPackageHandleStandardArgs.cmake:272 (message)
  cmake/Findzlib.cmake:39 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  src/CMakeLists.txt:13 (find_package)
```
This happens because the target must be `ZLIB`, with capital letters, and not `zlib`:

https://cmake.org/cmake/help/git-master/module/FindZLIB.html


fundamental:
Thanks for the fix :+1:

PR #64, opened 2020-06-30
Add sndio support

kinichiro:
This PR adds sndio support.
I had tested this on OpenBSD 6.7 (amd64), and saw both audio out and midi in.


fundamental:
Thanks for the functionality addition. If any bugs come up in this new sndio backend, you're up for resolving them, right?

kinichiro:
Thanks for merging.
And yes, I would like to see if bugs are reported.

PR #65, opened 2020-07-01
Universal improve cmake build

kinichiro:
Currently, cmake build appears to have some improvement points.

- It forces to link librt even if OS does not have it.
- "#ifdef HAVE_SCHEDULER" in Util.cpp does not work correctly,
  since cmake always defines "add_definitions(-DHAVE_SCHEDULER=" except Darwin.
- Some OS requires libX11 path to link.
- Only APPLE and FreeBSD can include C++ std::complex in globals.h

This PR solves these issues above.

- Check if OS has librt, and do not link it if not exist
- Do not define HAVE_SCHEDULER if OS does not have it
- Add libX11 path to link directory
- Check if OS has C++ std::complex, and use this condition to
  include <complex> or not in globals.h, instead of checking OS macro


kinichiro:
I will repost this with separate commit.

PR #67, opened 2020-07-08
[CMAKE] Remove duplicated check on PORTAUDIO.

carlo-bramini:


fundamental:
Thanks for the redundancy elimination :+1:

PR #68, opened 2020-07-09
[CYGWIN] Add support for this platform.

carlo-bramini:
When building under CYGWIN, I got few errors because some stuff, like `usleep()`, was not declared.
Actually, everything already exists in CYGWIN, but when compiling under C++11, some prototypes and symbols are intentionally deactivated and they need to be activated manually with some macros like `_GNU_SOURCE`, `_BSD_SOURCE` or others.
The addition of the required macros in this commit allows to activate the required symbols and compile the sources successfully.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -338,6 +338,10 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
     add_definitions(-DWIN32)
 endif()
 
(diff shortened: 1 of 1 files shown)


fundamental:
Neat, that's less changes than I expected. It's been a while since I've used cygwin (WSL appears to be slowly taking over that space), but adding a few flags for cygwin seems great :+1:

PR #69, opened 2020-07-11
[WIN32] Fix error on missing time functions.

carlo-bramini:
When building with mingw-w64, I suddently got this error:

```
[ 52%] Building CXX object src/CMakeFiles/zynaddsubfx_core.dir/Misc/Part.cpp.obj
src/Misc/Part.cpp: In lambda function:
src/Misc/Part.cpp:174:17: error: 'time' was not declared in this scope; did you mean 'ftime'?
  174 |                 time (&rawtime);
      |                 ^~~~
      |                 ftime
src/Misc/Part.cpp:175:45: error: 'localtime' was not declared in this scope
  175 |                 const struct tm* timeinfo = localtime (&rawtime);
      |                                             ^~~~~~~~~
src/Misc/Part.cpp:176:17: error: 'strftime' was not declared in this scope; did you mean 'ftime'?
  176 |                 strftime (p->loaded_file,23,"%F_%R.xiz",timeinfo);
      |                 ^~~~~~~~
      |                 ftime
make[2]: *** [src/CMakeFiles/zynaddsubfx_core.dir/build.make:447: src/CMakeFiles/zynaddsubfx_core.dir/Misc/Part.cpp.obj] Error 1
make[1]: *** [CMakeFiles/Makefile2:1659: src/CMakeFiles/zynaddsubfx_core.dir/all] Error 2

```
The solution is to include `ctime` near the other include files.
I tested it also with Debian and CYGWIN so it should be fine

### DIFF
--- a/src/Misc/Part.cpp
+++ b/src/Misc/Part.cpp
@@ -31,6 +31,7 @@
 #include <cstdio>
 #include <cstring>
 #include <cassert>
+#include <ctime>
 
 #include <rtosc/ports.h>
 #include <rtosc/port-sugar.h>


fundamental:
Yep, makes sense that there was a missing header since that's fairly new code to have a timecode savefile name fallback. I've been trying to get more platforms involved in the CI of zyn/rtosc/mruby-zest/etc to more easily catch these sorts of things.

Thanks for catching the issue :+1:

PR #70, opened 2020-07-11
[CMAKE] Fix detection of OSS

carlo-bramini:
When compiling for Windows, I discovered that CMake reported that OSS was found, which is not true since it is known to be unsupported by that platform.
I suggest to fix the included CMake script to avoid detection under WIN32 and also under APPLE.
This fix has been inspired by the FindOSS script implemented for GNUradio:

https://github.com/gnuradio/gnuradio/blob/master/cmake/Modules/FindOSS.cmake


fundamental:
Looks good to me :+1:

PR #71, opened 2020-07-12
Check if OS has librt, and do not link it if not exist

kinichiro:
Some OS doesn't have librt.
It should  link librt if OS has it.


fundamental:
Thanks for the addition :+1:

PR #72, opened 2020-07-12
Do not define HAVE_SCHEDULER if OS does not have it

kinichiro:
"#ifdef HAVE_SCHEDULER" in Util.cpp does not work correctly,
since cmake always defines "add_definitions(-DHAVE_SCHEDULER=" except Darwin.
It should not define HAVE_SCHEDULER if OS does not have it.


fundamental:
Seems reasonable.

PR #73, opened 2020-07-12
Add libX11 path to link directory

kinichiro:
Some OS requires libX11 path to link, since it is sometimes placed under /usr/X11R6/lib/ etc.
This PR adds libX11 path to link directory.


fundamental:
Ideally this should be handled by the find scripts for FLTK/NTK (older cmake find scripts have a few arcane variables they leave the info in), but this fix seems reasonable enough to me.

PR #74, opened 2020-07-12
Include <complex> other than APPLE and FreeBSD

kinichiro:
Currently, only APPLE and FreeBSD can include C++ std::complex in globals.h.
Check if OS has C++ std::complex, and use this condition to include or not in globals.h, instead of checking OS macro.


fundamental:
Looks good to me :+1:

PR #77, opened 2020-07-17
[WIN32] Fix problem of missing asprintf()

carlo-bramini:
Since existing implementation does not check if `asprintf()` is executed successfully, I left modified code also in the same manner, to be sure that new code works exactly like the old one. Probably, it should just return an error, but the value is unknown to me.


fundamental:
I'm fairly sure that no one utilizes the NSM codepath on windows, but since this is an equivalent implementation, no problem.
Thanks for removing the need for another #ifndef statement. :+1:

PR #78, opened 2020-09-12
Let MwDataObj::chain use a queue

JohannesLorenz:
This leads to chained messages being sent to the backend *after* they
are forwarded themselves.

### DIFF
--- a/src/Misc/MiddleWare.cpp
+++ b/src/Misc/MiddleWare.cpp
@@ -31,6 +31,7 @@
 #include "../UI/Fl_Osc_Interface.h"
 
 #include <map>
+#include <queue>
 
 #include "Util.h"
 #include "CallbackRepeater.h"
@@ -428,6 +429,9 @@ public:
 
 class MiddleWareImpl
 {
+    // messages chained with MwDataObj::chain
+    // must yet be handled after a previous handleMsg
+    std::queue<std::vector<char>> msgsToHandle;
 public:
     MiddleWare *parent;
     Config* const config;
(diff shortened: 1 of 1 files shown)


src/Misc/MiddleWare.cpp
```diff
@@ -714,6 +718,13 @@ class MiddleWareImpl
     // Handle an event with special cases
     void handleMsg(const char *msg);
 
+    // Add a message for handleMsg to a queue
+    void queueMsg(const char* msg)
+    {
+        msgsToHandle.emplace(msg, msg+rtosc_message_length(msg,
+                                          std::numeric_limits<size_t>::max()));
```
fundamental, line 725: If I'm not mistaken rtosc_message_length(msg, -1) also does this same behavior.

src/Misc/MiddleWare.cpp
```diff
@@ -2067,6 +2080,14 @@ void MiddleWareImpl::handleMsg(const char *msg)
     } else {
         //printf("Message Handled<%s:%s>...\n", msg, rtosc_argument_string(msg));
     }
+
+    // now handle all chained messages
+    while(!msgsToHandle.empty())
+    {
+        std::vector<char> front = msgsToHandle.front();
```
fundamental, line 2087: looks right, though I always have to check the docs since C++ is the only lang I've used where queue.pop() doesn't return the value that was popped.

fundamental:
Your call if you want to change the rtosc_message_length() call. Otherwise it looks good to merge. Ideally this would be something with higher test coverage, but I can't think of any situation where this sort of reordering would impact things so far.

src/Misc/MiddleWare.cpp
JohannesLorenz: `-1` is also used in other places, so I think it would look more consistent.

src/Misc/MiddleWare.cpp
JohannesLorenz: Yeah, I checked the `std::queue` docs carefully because I wasn't sure, too. But this one is right.

fundamental:
Should be good to merge. Let me know if you prefer me hitting the button.

JohannesLorenz:
I'll merge this now, untested. We'll may have tests soon, I think this will be OK.

PR #79, opened 2020-09-19
Bank damage

friedolino78:


fundamental:
Seems reasonable. I could imagine some consequences with the FLTK UI bugging out, but I think the way it accesses banks is sufficiently different that this shouldn't be a problem.

Thanks for the enhance. :+1:

PR #80, opened 2020-09-23
Plevel

friedolino78:
using the automate function i found out, that a few params cant be used as path.
tracked it down to an zyn error message about unknown bounds.
this patches sucessfully solved it for Plevel, Phpf and Plpf.


friedolino78:
pl redundant to the last commit in master.

PR #82, opened 2020-10-03
src/Nio: Fix missing CMake include when JACK is enabled

simonvanderveldt:
Small fix for a missing CMake include. Compilation would fail when building with only JACK enabled because `CheckIncludeFiles` wasn't included. It wouldn't fail when for example ALSA was enabled as well because within the CMake config for ALSA it would include `CheckIncludeFiles` which made it work for JACK as well, it's probably gone unnoticed so far because of this.


fundamental:
Good catch. With the first version of this module there wasn't metadata support, so this seems to be a detail overlooked when adding that functionality.

PR #83, opened 2020-10-07
Disable gcc optimizations for a specific function

justnope:
Due to a gcc opt bug a piece of code was put in a separate function to
prevent gcc from optimizing it. Later versions of gcc inlined the
function and the bug reappeared. So now we explicitly tell gcc to not
optimize it.

https://github.com/zynaddsubfx/zyn-fusion-build/issues/65


fundamental:
Thanks for the assist :+1:

PR #85, opened 2020-10-11
Ladder

friedolino78:
algorithm basically from discussion on:
https://www.kvraudio.com/forum/viewtopic.php?f=33&t=349859

### DIFF
--- a/src/DSP/MoogFilter.cpp
+++ b/src/DSP/MoogFilter.cpp
@@ -19,7 +19,7 @@ MoogFilter::MoogFilter(unsigned char Ftype, float Ffreq, float Fq,
     settype(Ftype); // q must be set before
     for (unsigned int i = 0; i<(sizeof(state)/sizeof(*state)); i++)
     {
-        state[i] = 0.0f;
+        state[i] = 0.0001f;
     }
 }
 
@@ -39,10 +39,14 @@ inline float MoogFilter::tanhX(const float x) const
 
 inline float MoogFilter::tanhXdivX(float x) const
 {
-    // Pade approximation for tanh(x)/x used in filter stages
+    
+    //add DC offset to raise even harmonics
+    x+= 0.01f;
     float x2 = x*x;
-    //~ return ((15.0+x2)/(15.0+6.0*x2)); // more accurate but instable at high frequencies
-    return (1.0f-(0.35f*x2)+(0.1f*x2*x2));
+    // Pade approximation for tanh(x)/x used in filter stages (5x per sample)
+    //~ return ((15.0+x2)/(15.0+6.0*x2)); 
+    // faster approximation without division
+    return (1.0f-(0.35f*x2)+(0.1f*x2*x2)); 
 }
 
 inline float MoogFilter::step(float input)
@@ -121,8 +125,8 @@ inline float MoogFilter::tan_2(const float x) const
 {
     //Pade approximation tan(x) hand tuned to map fCutoff
     float x2 = x*x;
-    //~ return ((9.54f*x*((11.08f - x2)))/(105.0f - x2*(45.0f + x2)));
-    return (x+0.15f*x2+0.3f*x2*x2);
+    //~ return ((9.54f*x*((11.08f - x2)))/(105.0f - x2*(45.0f + x2))); // more accurate but instable at high frequencies
+    return (x+0.15f*x2+0.3f*x2*x2); // faster, no division (by zero)
 }
 
 void MoogFilter::setfreq(float ff)
--- a/src/DSP/MoogFilter.cpp
+++ b/src/DSP/MoogFilter.cpp
@@ -1,0 +1,174 @@
+#include <cassert>
+#include <cstdio>
+#include <cmath>
+#include <stdio.h>
+
+#include "../Misc/Util.h"
+#include "MoogFilter.h"
+
+// theory from "THE ART OF VA FILTER DESIGN"
+// by Vadim Zavalishin
+
+namespace zyn{
+
+MoogFilter::MoogFilter(unsigned char Ftype, float Ffreq, float Fq,
+    unsigned int srate, int bufsize)
+    :Filter(srate, bufsize), sr(srate), gain(1.0f), type(Ftype)
+{
+    setfreq_and_q(Ffreq/srate, Fq);
+    settype(type); // q must be set before
+    for (unsigned int i = 0; i<(sizeof(state)/sizeof(*state)); i++)
+    {
+        state[i] = 0.0f;
+    }
+}
+
+MoogFilter::~MoogFilter(void)
+{
+
+}
+
+inline float MoogFilter::tan_2(const float x)
+{
+    //Pade approximation tan(x) hand tuned to map fCutoff
+    float x2 = x*x;
+    return ((9.5f*(11.15f*x - x2*x))/(105.0f - 45.0f*x2 + x2*x2));
+}
+
+inline float MoogFilter::tanhX(const float x)
+{
+    // Pade approximation of tanh(x) bound to [-1 .. +1]
+    // https://mathr.co.uk/blog/2017-09-06_approximating_hyperbolic_tangent.html
+    float x2 = x*x;
+    return (x*(105.0f+10.0f*x2)/(105.0f+(45.0f+x2)*x2)); //
+}
+
+
+inline float MoogFilter::tanhXdivX(float x)
+{
+    // Pade approximation for tanh(x)/x used in filter stages
+    float x2 = x*x;
+    return ((x2 + 105.0f)*x2 + 945.0f) / ((15.0f*x2 + 420.0f)*x2 + 945.0f);
+}
+
+inline float MoogFilter::step(float input)
+{
+    // transconductance
+    // gM(vIn) = tanh( vIn ) / vIn
+    float gm0 = tanhXdivX(state[0]);
+    float gm1 = tanhXdivX(state[1]);
+    float gm2 = tanhXdivX(state[2]);
+    float gm3 = tanhXdivX(state[3]);
+
+    // pre calc often used terms
+    float ctgm0 = c*gm0;
+    float ctgm1 = c*gm1;
+    float ctgm2 = c*gm2;
+    float ctgm3 = c*gm3;
+
+    // denominators
+    float d0 = 1.0f / (1.0f + ctgm0);
+    float d1 = 1.0f / (1.0f + ctgm1);
+    float d2 = 1.0f / (1.0f + ctgm2);
+    float d3 = 1.0f / (1.0f + ctgm3);
+
+    // pre calc often used term
+    float gm1td2tgm2td3 = gm1 * d2 * gm2 * d3;
+
+    // instantaneous response estimate
+    float y3Estimate =
+        cp4 * d0 * gm0 * d1 * gm1td2tgm2td3 * input +
+        cp3 * gm0 * d1 * gm1td2tgm2td3      * d0 * state[0] +
+        cp2 * gm1td2tgm2td3                 * d1 * state[1] +
+        c   * gm2 * d3                      * d2 * state[2] +
+                                              d3 * state[3];
+
+    // instantaneous gain coefficient
+    float z0 = ctgm0 * d0;
+    float z1 = ctgm1 * d1;
+    float z2 = ctgm2 * d2;
+    float z3 = ctgm3 * d3;
+    float instantaneousgain = 1.0f / (1.0f + feedbackGain * z0*z1*z2*z3);
+
+    // input for the fist stage
+    float u = input - tanhX(feedbackGain * y3Estimate) * instantaneousgain;
+    // output of all stages
+    float y0 = gm0 * d0 * (state[0] + c * u);
+    float y1 = gm1 * d1 * (state[1] + c * y0);
+    float y2 = gm2 * d2 * (state[2] + c * y1);
+    float y3 = gm3 * d3 * (state[3] + c * y2);
+
+    // update state
+    state[0] += ct2 * (u - y0);
+    state[1] += ct2 * (y0 - y1);
+    state[2] += ct2 * (y1 - y2);
+    state[3] += ct2 * (y2 - y3);
+
+    // calculate multimode filter output
+    return (a0 * u
+          + a1 * y0
+          + a2 * y1
+          + a3 * y2
+          + a4 * y3);
+}
+
+void MoogFilter::filterout(float *smp)
+{
+    for (int i = 0; i < buffersize; i ++)
+    {
+        smp[i] = step(tanhX(smp[i]*gain));
+        smp[i] *= outgain;
+    }
+}
+
+void MoogFilter::setfreq_and_q(float frequency, float q_)
+{
+    setfreq(frequency/sr);
+    setq(q_);
+}
+
+void MoogFilter::setfreq(float ff)
+{
+    // limit cutoff to prevent overflow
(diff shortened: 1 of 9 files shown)


friedolino78:
the last commit has broken something. working on it

friedolino78:
still some unexpected overflows

friedolino78:
3rd attempt. I hope it is stable now.

src/DSP/MoogFilter.h
```diff
@@ -0,0 +1,52 @@
...
+        unsigned char type;
+        
+        float step(float x);
+
+        float tanhXdivX(const float x);
+        float tanhX(const float x);
+        float tan_2(const float x);
+        
+        float fb;
+        float a0, a1, a2, a3, a4;
```
JohannesLorenz, line 45: Most of these variable names don't make clear what they are used for. Though, you could argue that you just need to look up the cpp file to find out what they mean.

src/DSP/MoogFilter.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+namespace zyn{
+
+MoogFilter::MoogFilter(unsigned char Ftype, float Ffreq, float Fq,
+    unsigned int srate, int bufsize)
+    :Filter(srate, bufsize), sr(srate), gain(1.0f), type(Ftype)
+{
+    setfreq_and_q(Ffreq/srate, Fq);
+    settype(type); // q must be set before
+    for (int i = 0; i>4; i++)
```
JohannesLorenz, line 20: Wait, do I read this right? :smile:

src/DSP/MoogFilter.cpp
```diff
@@ -0,0 +1,174 @@
...
+                                              d3 * state[3];
+
+    // instantaneous gain coefficient
+    float z0 = ctgm0 * d0;
+    float z1 = ctgm1 * d1;
+    float z2 = ctgm2 * d2;
+    float z3 = ctgm3 * d3;
+    float fbgain = 1.0f / (1.0f + fb * z0*z1*z2*z3);
+
+    // input for the fist stage
```
JohannesLorenz, line 93:
```suggestion
    // input for the first stage
```

JohannesLorenz (commented):
Incomplete review. I think someone with more DSP knowledge must review it.

I left some comments as far as I understood the code. Mostly minor things, hopefully it helps.

src/DSP/MoogFilter.cpp
```diff
@@ -0,0 +1,174 @@
...
+}
+
+MoogFilter::~MoogFilter(void)
+{
+
+}
+
+inline float MoogFilter::tan_2(const float x)
+{    
+    //Pade approximation tan(x) hand tuned to map fCutoff
```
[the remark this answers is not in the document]
friedolino78: > This seems to be mostly the same as `tanhX`.
> 
> 1. Did you really mean to approximate `tan` instead of `tanh`?
> 2. Is it intended that `45.0f * x2` is _subtracted_ here instead of added as in `tanhX`?
> 3. How did you get the "hand tuning" done - any source, or did you do it on your own? I guess you there is already literature on how to do it?

1+2. it is intended because it is tan, not tanh ;-)
3. hand tuning was made by measuring resonance frequency vs the one set via osc.

fundamental:
Good to see that this PR is still alive :) Let me know when you'd like a review. This one is a bit more involved, but the VA filter guide isn't so complex that it can't get done.

friedolino78:
> Good to see that this PR is still alive :) Let me know when you'd like a review. This one is a bit more involved, but the VA filter guide isn't so complex that it can't get done.

Just a rebase to current Master. It's ready for review

src/DSP/MoogFilter.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+void MoogFilter::setgain(float dBgain)
+{
+    gain = dB2rap(dBgain);
+}
+
+void MoogFilter::settype(unsigned char type_)
+{
+    type = type_;
+    switch (type)
```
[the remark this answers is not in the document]
JohannesLorenz: @friedolino78 can you please fix this?

fundamental:
Stability and default gain is a good bit better than I expected. One minor thing is that it appears the FLTK GUI was updated, however the category selection works correctly, but right now the filter type is only LPF which misses out on the two other types. I think changing via oscprompt has worked so far. I'll update with more comments as I work though the functionality.

src/DSP/MoogFilter.h
```diff
@@ -0,0 +1,53 @@
...
+        ~MoogFilter() override;
+        void filterout(float *smp) override;
+        void setfreq(float /*frequency*/) override;
+        void setfreq_and_q(float frequency, float q_) override;
+        void setq(float /*q_*/) override;
+        void setgain(float dBgain) override;
+        void settype(unsigned char type); //
+
+    private:
+        unsigned sr;
```
fundamental, line 34: sr and gain seem to be redundant given that Filter has 'samplerate' and 'outgain'.

src/DSP/MoogFilter.h
```diff
@@ -0,0 +1,53 @@
...
+        //! @param Fq resonance, range [0.1,1000], logscale
+        MoogFilter(unsigned char Ftype, float Ffreq, float Fq,
+                unsigned int srate, int bufsize);
+        ~MoogFilter() override;
+        void filterout(float *smp) override;
+        void setfreq(float /*frequency*/) override;
+        void setfreq_and_q(float frequency, float q_) override;
+        void setq(float /*q_*/) override;
+        void setgain(float dBgain) override;
+        void settype(unsigned char type); //
```
fundamental, line 31: start of a comment without any comment?

src/DSP/MoogFilter.h
```diff
@@ -0,0 +1,53 @@
...
+    private:
+        unsigned sr;
+        float gain;
+        unsigned char type;
+
+        float step(float x);
+
+        float tanhXdivX(const float x);
+        float tanhX(const float x);
+        float tan_2(const float x);
```
fundamental, line 42: Somewhat pedantic, so feel free to ignore, but if possible marking functions const is handy in case we're trying to get info like filter response from a const object. Not really an issue though.

src/Params/FilterParams.cpp
```diff
@@ -131,7 +131,9 @@ const rtosc::Ports FilterParams::ports = {
     {"type-svf::i", rProp(parameter) rShort("type")
         rOptions(low, high, band, notch)
             rDoc("Filter Type"), 0, rOptionCb(Ptype)},
-
+    {"type-moog::i", rProp(parameter) rShort("type")
+        rOptions(HP, BP, LP)
+            rDoc("Filter Type"), 0, rOptionCb(Ptype)},
```
fundamental, line 136: looks good

src/DSP/MoogFilter.cpp
```diff
@@ -0,0 +1,174 @@
...
+    {
+        case 1:
+            a0 = 0.0f; a1 = 0.0f; a2 = 4.0f; a3 =-8.0f; a4 = 4.0f;
+            break;
+        case 0:
```
fundamental, line 164: Silly request, but putting the cases in order makes it a bit easier to read.

src/DSP/MoogFilter.h
fundamental: Ignore the gain comment. It appears that gain and outgain are semantically different if my read of the rest of the code is right.

src/DSP/MoogFilter.h
```diff
@@ -0,0 +1,53 @@
...
+        void setfreq(float /*frequency*/) override;
+        void setfreq_and_q(float frequency, float q_) override;
+        void setq(float /*q_*/) override;
+        void setgain(float dBgain) override;
+        void settype(unsigned char type); //
+
+    private:
+        unsigned sr;
+        float gain;
+        unsigned char type;
```
fundamental, line 36: Type appears to be a write only variable. The only place I see it used is in setType(). So, you should be able to remove it without a problem. Pretty minor all things considered as saving a byte in a full zyn instance is basically nothing :p

fundamental:
Seems good for the most part. From my read the code should optimize pretty well (mostly 4x of the same operation should let gcc get everything into SSE2 ops). There's minor issues here and there documented which should be easy enough for you to fix pre-merge. There was one pop where there was a volume spike. I might try to replicate, but otherwise things seem pretty close to mergable.

src/DSP/MoogFilter.h
friedolino78: > Ignore the gain comment. It appears that gain and outgain are semantically different if my read of the rest of the code is right.

yes, (input) gain drives the nonlinerarity. it's a sonically important feature

src/DSP/MoogFilter.h
friedolino78: it doesnt seem to care.
zynaddsubfx-code/src/DSP/MoogFilter.h:40:9: warning: type qualifiers ignored on function return type [-Wignored-qualifiers]
40 |         const float tanhXdivX(const float x);

looks like in C11 mode, type qualifiers are completely ignored on function return types

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=39985#c5
https://gcc.gnu.org/git/?p=gcc.git&a=commit;h=aa4b467b680f230ab11922d1e29695e1eaba12af

fundamental: To be specific I mean "float tanhX(float x) const" not "const float tanhX(float x)" or "float tanhX(const float x)"

src/DSP/MoogFilter.cpp
```diff
@@ -0,0 +1,174 @@
...
+    ct2 = c * 2.0f;
+    cp2 = c * c;
+    cp3 = cp2 * c;
+    cp4 = cp2 * cp2;
+}
+
+void MoogFilter::setq(float q)
+{
+    // flattening the Q input
+    feedbackGain = cbrtf(q/1000.0f)*4.0f + 0.1f;
```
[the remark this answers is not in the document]
friedolino78: that's the good thing of not knowing anything and being forced to permanently google and keep references open ;-)

src/DSP/MoogFilter.h
friedolino78: ok

src/DSP/MoogFilter.h
friedolino78: :-)

src/DSP/MoogFilter.cpp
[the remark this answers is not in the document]
friedolino78: fixed.

fundamental:
Adjusting the setQ() routine to allow for more agressive resonance allows for us to see working self oscillation. There's a few sidelobes, but that's likely just the non-linearity approximations. After adjusting the range I was able to see self resonance at feedbackGain above about 4.125 with the frequency centered on the knob (roughly).

I'd recommend replicating the result I observed and then seeing about if my hypothesis about the non-linearities is correct. For the record I never measured non-base harmonic content on my physical analog ladder filters, but I suspect any sidelobes are much smaller.

Also of note, thanks for adding in the LP/BP/HP options for the moog filter. I'm seeing some inconsistency in them though. That might be that they're labeled out of order or more likely in the fltk GUI there's a refresh() method which might need to re-fetch that parameter when changing into a moog filter.

friedolino78:
> Adjusting the setQ() routine to allow for more agressive resonance allows for us to see working self oscillation. There's a few sidelobes, but that's likely just the non-linearity approximations. After adjusting the range I was able to see self resonance at feedbackGain above about 4.125 with the frequency centered on the knob (roughly).
> 

I cannot reproduce your Q Problem. here it starts oscillating even with only DC excitation with feedbackGain>4.008
With this value it takes about 5 seconds and a third harmonic sidelobe is with -120dB baraly noticable in the spectrogram if you know where to look. With feedbackGain's current max 4.1 it takes 0.5 seconds to full oscillation. the 3rd harmonic has -90dB.

with feedback > 4 the stages are driven more into the nonlinear part tanh approximation  for "input saturation".
maybe we could make it more gentle (even harmonics) by adding some DC offset

friedolino78:
interesting observations: 
with feedbackGain: 4.005953 for lower freqs it takes about 20 seconds until the self oscillation slowly rises.
for higher freqs it works faster.

with an offset of 0.01 inside the tanhx/x approximation the third harmonic gets accompanied by a second harmonic. unfortunatelly i can't tell the difference acoustically.

fundamental:
I discarded my changes while working on the build system update, but I
was able to see when the filter was self oscillating there were side
lobes, not just harmonics.

I'm guessing some pictures would help move the discussion forward a bit.
You use baudline for spectral analysis, right?

friedolino78:
> I discarded my changes while working on the build system update, but I was able to see when the filter was self oscillating there were side lobes, not just harmonics. I'm guessing some pictures would help move the discussion forward a bit. You use baudline for spectral analysis, right?

seems like i had a different meaning in mind when talking about sidelobe.

what i noticed were aliases symmetrical to fNyquist/2. Is that what you meant?

fundamental:
Hm, interesting. I couldn't quite replicate the problem I was seeing. I did spot the harmonics you were talking about though. Experimenting around with different approximations it looks like they can be reduced a fair amount. Though of note, you initialize the state matrix with zeros. When using tanh(x)/x instead of an approximation that results in a load of NaNs (or +inf? if you don't have -ffast-math). Initializing those with a small non-zero value seems to be sufficient to avoid the problem. I just used 1e-3 for all of them, but it's unclear if there's a better value. Not sure if that complicated any of your past testing or not.

friedolino78:
> Hm, interesting. I couldn't quite replicate the problem I was seeing. I did spot the harmonics you were talking about though. Experimenting around with different approximations it looks like they can be reduced a fair amount. Though of note, you initialize the state matrix with zeros. When using tanh(x)/x instead of an approximation that results in a load of NaNs (or +inf? if you don't have -ffast-math). Initializing those with a small non-zero value seems to be sufficient to avoid the problem. I just used 1e-3 for all of them, but it's unclear if there's a better value. Not sure if that complicated any of your past testing or not.

the non zero initialization is a good idea. the tanh (approximation) is the single most cpu intensive line.
 it's a tradeoff. if you want it more linear, you can still reduce the input gain.

what do you think of the idea of  offsetting the tanh function to raise the even harmonic?
can you measure the harmonics of the hw filter, when slightly driven into saturation?

fundamental:
As requested: https://fundamental-code.com/tmp/2021-06-synth.wav https://fundamental-code.com/tmp/2021-06-synth2.wav I had some better takes, but something was wrong with the recording sample rate on my sound input until I restarted my computer and wasted a lot of time recording broken inputs. Also, it's a lot harder to setup something for recording when I'm not listening to the source.

fundamental:
Also FYI the self resonance starts at about 75% resonance on the physical slider on the VCF.

friedolino78:
> As requested: https://fundamental-code.com/tmp/2021-06-synth.wav https://fundamental-code.com/tmp/2021-06-synth2.wav I had some better takes, but something was wrong with the recording sample rate on my sound input until I restarted my computer and wasted a lot of time recording broken inputs. Also, it's a lot harder to setup something for recording when I'm not listening to the source.

I made spectrograms of your record and zyn with x offset 0.01 side by side: https://imgur.com/a/QoE6WVy
the amount of distortion is 20dB higher in your record. So maybe i can tweak the approximation a bit.

fundamental:
FYI, the other audio samples are pending getting a cheap audio input device. After multiple reboots I've got no idea what is causing the audio line-in recording on my desktop to be inconsistent.

friedolino78:
i guess this one is quite ready.

PR #86, opened 2020-10-17
make ENVs repeatable

friedolino78:
New Feature: Repeating ENVs. Only a few lines of code, but lots of fun to play with.

### DIFF
--- a/src/Params/EnvelopeParams.cpp
+++ b/src/Params/EnvelopeParams.cpp
@@ -111,8 +111,11 @@ static const rtosc::Ports localPorts = {
             rPresetAtMulti(true, ad_global_amp, ad_global_filter, ad_voice_amp,
                                  ad_voice_fm_amp),
             rDefault(false),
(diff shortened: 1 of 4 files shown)


friedolino78:
Zest Counterpart is on:
[https://github.com/mruby-zest/mruby-zest/pull/43](https://github.com/mruby-zest/mruby-zest/pull/43)

fundamental:
This is a bit of functionality which I've thought about in the past. If desired it basically grants users the ability to have a custom LFO without adding complexity (that I don't think could reasonably fit into the LFO UX). I think this is something that should be advertised heavily in next release with the disclaimer that if people come up with different ways of interacting with the functionality that there may be some slack for us devs to break patches with the new functionality for the subsequent release or two.

friedolino78:
With your words in mind while cleaning up my local branches i just found the same thing in a branch from 2015.
In this branch I saw, something I forgot here: storing/loading this bool in XML
I added it in:
https://github.com/zynaddsubfx/zynaddsubfx/pull/87

PR #88, opened 2020-10-20
add MIDI clock handling

friedolino78:


friedolino78:
one unsolved problem is to find a good reference time for the phase.
right now the reference is the beginning of the first jack buffer.
So its only a relative the phase offset of the clock ticks.

What's needed would be the same common reference, that the other receiver of the clock signal use.
Maybe Timcode and Transport Messages are the answer.

fundamental:
Looks like things are progressing as expected. For jack MIDI + jack audio it looks like there's effort being put into keeping the sample offsets. For alsa->jack or other configurations (even alsa->alsa based upon our usage IIRC), I'm guessing we'd have to use logic similar to a2jmidid's codebase to establish a linkage between midi time and audio time.

Of minor note, it looks like there's a conflict in src/Params/LFOParams.cpp, not really an issue but something to resolve before finishing the feature.

friedolino78:
> I'm guessing we'd have to use logic similar to a2jmidid's codebase to establish a linkage between midi time and audio time.
for alsa->jack the key in a2jmidid seems to be the call to 
`jack_nframes_t | jack_frame_time (const jack_client_t *)`
-> we get a dependency to jack inside the alsa midi thread.

src/Nio/MidiIn.cpp
```diff
@@ -147,7 +170,96 @@ void MidiIn::midiProcess(unsigned char head,
...
+            break;
+        case 0xF8: /* M_CLOCK */
+            ev.type  = M_CLOCK;
+            InMgr::getInstance().putEvent(ev);
+            break;
+        case 0xFA: /* M_START */
+            ev.type  = M_START;
+            InMgr::getInstance().putEvent(ev);
+            break;
+        case 0xFB: /* M_STOP */
```
falkTX, line 253: small note here, 0xFB is continue, and 0xFC is stop.
you have it in reverse.

friedolino78: > small note here, 0xFB is continue, and 0xFC is stop.
> you have it in reverse.

thx

friedolino78:
LFO Tempo sync to global BPM value is working now in zest. it is possible to set speedratio via knob or set to fixed values with small integer rationals via selector box.
bpm can be set via OSC: 
oscsend localhost 7777 /bpm i 80
default is 120 bpm

src/Synth/LFO.cpp
```diff
@@ -50,7 +56,10 @@ LFO::LFO(const LFOParams &lfopars, float basefreq, const AbsTime &t, WatchManage
     else {
         const float tmp = fmod(t.time() * phaseInc, 1.0f);
         phase = fmod((lfopars.Pstartphase - 64.0f) / 127.0f + 1.0f + tmp, 1.0f);
```
friedolino78, line 58: somehow the global sync doesn't seem to work anymore despite the line being executed - proven by printf

fundamental:
Just making sure that I'm not mixing up PRs, this was broken into pieces with PR #117 being a prereq to this PR, right?

friedolino78:
> Just making sure that I'm not mixing up PRs, this was broken into pieces with PR #117 being a prereq to this PR, right?

correct. this one will be rebased after 117 is merged.

friedolino78:
> Just making sure that I'm not mixing up PRs, this was broken into pieces with PR #117 being a prereq to this PR, right?

exact one year later this branch is rebased.

friedolino78:
next step will be to make its functionality to match the behaviour described in:
[http://midi.teragonaudio.com/tech/midispec/seq.htm](http://midi.teragonaudio.com/tech/midispec/seq.htm)

PR #90, opened 2020-10-21
Implement workaround for old presets.

hselasky:
All effect parameters have a default value.
Default values are skipped when storing parameters,
and must appear as the default value when loading.

Up until recently it was assumed that the default
value of all parameters is zero. This is no longer
true, but when loading old presets we need to
preserve this behaviour! Else sounds may change.

Add exception for the recently added DC offset parameter.

Bump the ZynAddSubFX version.

Signed-off-by: Hans Petter Selasky <hps@selasky.org>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/")
 project(zynaddsubfx)
 set(VERSION_MAJOR "3")
 set(VERSION_MINOR "0")
-set(VERSION_REVISION "5")
+set(VERSION_REVISION "6")
 
 #Set data directory, if any
 if(DEFINED ZYN_DATADIR)
--- a/src/Effects/EffectMgr.cpp
+++ b/src/Effects/EffectMgr.cpp
@@ -502,7 +502,25 @@ void EffectMgr::getfromXML(XMLwrapper& xml)
     if(xml.enterbranch("EFFECT_PARAMETERS")) {
         for(int n = 0; n != 128; n++) {
             if(xml.enterbranch("par_no", n) == 0) {
-                settings[n] = -1; /* use default */
+                /*
+                 * XXX workaround for old presets:
+                 *
+                 * All effect parameters have a default value.
+                 * Default values are skipped when storing parameters,
+                 * and must appear as the default value when loading.
+                 *
+                 * Up until recently it was assumed that the default
+                 * value of all parameters is zero. This is no longer
+                 * true, but when loading old presets we need to
+                 * preserve this behaviour! Else sounds may change.
+                 */
+                if (xml.fileversion() < version_type(3,0,6) &&
+                    /* XXX old presets don't have DC offset */
+                    (geteffect() != 6 || n < 11)) {
+                        settings[n] = 0;
+                } else {
+                        settings[n] = -1; /* use parameter default */
+                }
             } else {
                 settings[n] = xml.getpar127("par", 0);
                 xml.exitbranch();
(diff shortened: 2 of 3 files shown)


hselasky:
This commit synchronizes the GITHUB code with the sourceforge GIT repository.

JohannesLorenz:
@hselasky Do you have an idea why `PluginTest` failed?

JohannesLorenz:
From what I understand, if you want to sync sf and gh, you need a commit on top of both with the same ID. That can only be done by merging. This is not a merge commit, so it will not sync sf with gh.

This commit only cherry-picks the first diverging sf commit to gh.

hselasky:
> @hselasky Do you have an idea why `PluginTest` failed?

No, do we have a log somewhere?

JohannesLorenz:
> No, do we have a log somewhere?

Click the red cross on this page. (There are two, one at the bottom and one next to your commit ID)

hselasky:
> From what I understand, if you want to sync sf and gh, you need a commit on top of both with the same ID. That can only be done by merging. This is not a merge commit, so it will not sync sf with gh.
> 
> This commit only cherry-picks the first diverging sf commit to gh.

I see. This is not a cherry-pick. It is a full rebase, and the missing patch has been put on top after rebase.

I'm not so good with merging. I thought the order of commits doesn't matter that much?

--HPS

hselasky:
```
In PluginTest::testLoadSave:

/home/travis/build/zynaddsubfx/zynaddsubfx/build/src/Tests/../../../src/Tests/PluginTest.h:234: Error: Expected ((int)(fdata.length()+1) == res), found (149605 != 155500)

/home/travis/build/zynaddsubfx/zynaddsubfx/build/src/Tests/../../../src/Tests/PluginTest.h:235: Error: Assertion failed: fdata == result

3898 vs 4213

[WARNING] XML File appears to be radically different

Failed 1 and Skipped 0 of 3 tests
```

I need to update those asserts. Let me fix that.

hselasky:
I found that the test guitar file needs to be re-generated. Let's hope that the build passes all tests now!

fundamental:
That commit doesn't look right, there's a huge number of changes to the test .xmz file involving floating point values which should be bit-per-bit identical.

fundamental:
Pulling the current github HEAD, it looks like PluginTest is passing, so I would expect the only changes to be the additional "effect.par = 0" semantics to be added by 4c3e1b8. The code I wrote in that test has a basic DIY 2-way diff algorithm to print out which things are different and it gives up if it looks like too many lines differ. I guess it's time to grab 4c3e1b8 and see what lines are changed on my system.

fundamental:
I've attached the diff that I can observe on my system which appears to be much smaller than the one proposed in 74e72ba

Diff in question: [test-guitar.diff](https://github.com/zynaddsubfx/zynaddsubfx/files/5431875/test-guitar.diff.txt)

Side note, do you want to have access to the github repo? It's an oversight if you don't currently have commit access given that you have sourceforge commit access.

hselasky:
Yes, please, having access at GitHub too would be nice.

hselasky:
@fundamental : Can you try compiling with clang ? I agree that the difference is very strange, and maybe it is something related to my environment.

fundamental:
I can confirm that the above diff works with gcc 9.3.0 and clang 10.0.1 on my Slackware-current x86_64 system.
This year I've done some hardware upgrade specifically to enable easier VM/docker testing/development for zyn. While I still find VMs pretty cumbersome, I have had quite a bit of success with docker. Let me know if there's a target version of one or the other and some linux distro base image you'd want me to run locally.

I wouldn't be all that surprised if there's another compiler bug to be worked around, though if there is one then it should be a much more minimal test case than the last optimization glitch.

hselasky:
OK, I updated the pull request.

fundamental:
Looks good to me. I'm seeing the same commits, but with different SHA ids over on sourceforge, I've merged the two to avoid force pushing either. We likely should start to deprecate the sourceforge version (using it as a mirror of the github rather than an independent remote) to avoid issues in the future.

PR #92, opened 2020-11-08
path-search/save_osc: include non-RT ports

JohannesLorenz:
Split MW snoop ports into non-RT ports and other snoop ports, and then
    
* Let path-search iterate over non-RT ports, too (e.g. to let port-checker see those ports)
* Let save_osc save non-RT ports, too (to build complete savefiles)

This commit is a bug fix, because currently, port-checker does not look at the non-RT MW-ports, and it would make the SaveOsc tests fail if the test xmz would have non-RT ports changed (in comparison to the default values).

This commit is preparation for the wavetable-branch (SaveOsc failed there with the new non-RT ports).

### DIFF
--- a/src/Effects/DynamicFilter.cpp
+++ b/src/Effects/DynamicFilter.cpp
@@ -27,7 +27,7 @@ namespace zyn {
 #define rEnd }
 
 rtosc::Ports DynamicFilter::ports = {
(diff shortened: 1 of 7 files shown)


JohannesLorenz:
I need to close this, I used the wrong master base.

PR #93, opened 2020-11-08
path-search/save_osc: include non-RT ports

JohannesLorenz:
Split MW snoop ports into non-RT ports and other snoop ports, and then

* Let path-search iterate over non-RT ports, too (e.g. to let port-checker see those ports)
* Let save_osc save non-RT ports, too (to build complete savefiles)

This commit is a bug fix, because currently, port-checker does not look at the non-RT MW-ports, and it would make the SaveOsc tests fail if the test xmz would have non-RT ports changed (in comparison to the default values).

This commit is preparation for the wavetable-branch (SaveOsc failed there with the new non-RT ports).

### DIFF
--- a/src/Misc/Master.cpp
+++ b/src/Misc/Master.cpp
@@ -388,6 +388,7 @@ static const Ports master_ports = {
     rRecursp(part, 16, "Part"),//NUM_MIDI_PARTS
     rRecursp(sysefx, 4, "System Effect"),//NUM_SYS_EFX
     rRecursp(insefx, 8, "Insertion Effect"),//NUM_INS_EFX
+    rRecur(HDDRecorder, "HDD recorder"),
     rRecur(microtonal, "Microtonal Mapping Functionality"),
     rRecur(ctl, "Controller"),
     rArrayOption(Pinsparts, NUM_INS_EFX, rOpt(-2, Master), rOpt(-1, Off),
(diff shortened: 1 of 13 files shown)


JohannesLorenz:
Funny behavior of oscprompt:

1. Before this commit, entering "/part" just brings "/part#16", and for "/part0/kit0/adpars/VoicePar1/OscilSmp/" all OscilGen ports (RT, non) are displayed.
2. With this commit, entering "/part" just brings "/part#16" and "/part#16/kit#16/adpars/VoicePar#8/OscilSmp/" (too much info), and for "/part0/kit0/adpars/VoicePar1/OscilSmp/" all OscilGen ports (RT, non) are displayed.
3. With this commit on the wavetable branch (where OscilSmp are only reachable via MW), entering "/part" just brings "/part#16" and "/part#16/kit#16/adpars/VoicePar#8/OscilSmp/" (too much info), and for "/part0/kit0/adpars/VoicePar1/O", OscilGen is not reachable anymore.

So this commit makes all worse for oscprompt :smile: However, I think these are bugs in oscprompt, not in this commit.

fundamental:
Per the behavior observed in oscprompt I'd say it's mostly a bug in rtosc::path_search than anything else. (At least I think that's the code which should simplify things as oscprompt should never really care about implementation details on the end of a remote client). Most of the code changes are in relation to osc save files which I personally haven't made too much use out of (besides the rather useful regression tests which have caught a few bugs). I will say that with the new middleware merged ports we might want to consider adding a hook for the osc-doc file output to use the new merge ports as well (though I'm guessing that will trigger a bug or two in that chunk of code). It's been a while since I've regenerated the html documentation for the OSC API, but usually I refer users over to that with respect to "what OSC port does X". Having live completion for middleware ports in oscprompt will be good, though presumably quirky for the time being.

JohannesLorenz:
Locally (in my local repo), I fixed the duplicate OscilSmp in oscprompt, and both oscdoc and json are fixed now.

@fundamental Which ports do you think of with "Having live completion for middleware ports in oscprompt will be good"?

fundamental:
Everything that you'd expect a user to want to probe. As it stands, most of
them are covered, but bank ports and saving/loading are notable exceptions.
Looking at the source, there's some stuff for IO, config, and undo/redo in
there, which might be good.

PR #95, opened 2020-11-22
Message fixes

JohannesLorenz:
Probably trivial to review. :smile:

### DIFF
--- a/src/Misc/MiddleWare.cpp
+++ b/src/Misc/MiddleWare.cpp
@@ -889,7 +889,7 @@ public:
     void kitEnable(int part, int kit, int type);
 
     // Handle an event with special cases
-    void handleMsg(const char *msg);
+    void handleMsg(const char *msg, bool msg_comes_from_realtime = false);
 
     // Add a message for handleMsg to a queue
     void queueMsg(const char* msg)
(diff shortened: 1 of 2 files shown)


src/Misc/MiddleWare.cpp
```diff
@@ -2265,10 +2265,16 @@ void MiddleWareImpl::handleMsg(const char *msg)
 
     //A message unmodified by snooping
     if(d.matches == 0 || d.forwarded) {
-        //if(strcmp("/get-vu", msg)) {
-        //    printf("Message Continuing on<%s:%s>...\n", msg, rtosc_argument_string(msg));
-        //}
-        uToB->raw_write(msg);
+        if(msg_comes_from_realtime) {
+            // don't reply the same msg to realtime - avoid cycles
+            printf("Message from RT will not be replied to RT: <%s:%s>...\n",
```
JohannesLorenz, line 2270: I'm going to comment that `printf` out yet. Was just for debugging.

fundamental:
Seems to match what was described in IRC, so we should be good to merge. Looks like travis-ci was acting up on the one build, so the one failure looks safe to ignore.

PR #96, opened 2020-11-30
fix default values for ENV time params

friedolino78:
problem with seems to be solved.

### DIFF
--- a/src/Params/EnvelopeParams.cpp
+++ b/src/Params/EnvelopeParams.cpp
@@ -118,11 +118,11 @@ static const rtosc::Ports localPorts = {
             "Repeat the Envelope"),
     rParamDT(A_dt ,  rShort("a.dt"), rLinear(0,127), "Attack Time"),
     rParamF(A_dt,  rShort("a.dt"), rLog(0.0f,41.0f), rDefaultDepends(loc),
-              rPreset(ad_global_freq, 0.254),   rPreset(ad_global_filter, 0.127),
-              rPreset(ad_voice_freq, 0.127),    rPreset(ad_voice_filter, 0.97),
-              rPreset(ad_voice_fm_freq, 3.62), rPreset(ad_voice_fm_amp, 1.876),
-              rPreset(sub_freq, 0.254),         rPreset(sub_bandwidth, 0.97),
-              rDefault(0.00),
+              rPreset(ad_global_freq, 0.254f),   rPreset(ad_global_filter, 0.127f),
+              rPreset(ad_voice_freq, 0.127f),    rPreset(ad_voice_filter, 0.970f),
+              rPreset(ad_voice_fm_freq, 3.620f), rPreset(ad_voice_fm_amp, 1.876f),
+              rPreset(sub_freq, 0.254f),         rPreset(sub_bandwidth, 0.970f),
+              rDefault(0.0f),
               "Attack Time"),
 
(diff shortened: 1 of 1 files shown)


src/Params/EnvelopeParams.cpp
```diff
@@ -324,17 +324,17 @@ void EnvelopeParams::init(zyn::consumer_location_t _loc)
...
-        case ad_voice_fm_amp:  ADSRinit(80, 90, 127, 100); break;
-        case sub_freq:         ASRinit(30, 0.025, 64, 0.5); break;
-        case sub_bandwidth:    ASRinit_bw(100, 0.97, 64, 0.5); break;
+        case sub_filter:       ADSRinit_filter(64, 0.127f, 64, 0.970f, 0.499f, 64); break;
+        case ad_voice_amp:     ADSRinit_dB(0, 6.978f, 127, 6.978f); break;
+        case ad_voice_freq:    ASRinit(30, 0.127f, 64, 0.499f); break;
+        case ad_voice_filter:  ADSRinit_filter(90, 0.970f, 40, 0.970f, 0.009f, 40); break;
+        case ad_voice_fm_freq: ASRinit(20, 3.620f, 40, 1.876f); break;
+        case ad_voice_fm_amp:  ADSRinit(1.876f, 3.620f, 127, 6.978f); break;
+        case sub_freq:         ASRinit(0.061f, 0.254f, 64, 0.499f); break;
```
JohannesLorenz, line 336: The first parameter expects a `char` here.

src/Params/EnvelopeParams.cpp
```diff
@@ -324,17 +324,17 @@ void EnvelopeParams::init(zyn::consumer_location_t _loc)
...
-        case sub_filter:       ADSRinit_filter(64, 0.01, 64, 0.025, 0.05, 64); break;
-        case ad_voice_amp:     ADSRinit_dB(0, 7.0, 127, 7.0); break;
-        case ad_voice_freq:    ASRinit(30, 0.01, 64, 0.5); break;
-        case ad_voice_filter:  ADSRinit_filter(90, 0.025, 40, 0.025, 0.009, 40); break;
-        case ad_voice_fm_freq: ASRinit(20, 3.62, 40, 1.876); break;
-        case ad_voice_fm_amp:  ADSRinit(80, 90, 127, 100); break;
-        case sub_freq:         ASRinit(30, 0.025, 64, 0.5); break;
-        case sub_bandwidth:    ASRinit_bw(100, 0.97, 64, 0.5); break;
+        case sub_filter:       ADSRinit_filter(64, 0.127f, 64, 0.970f, 0.499f, 64); break;
+        case ad_voice_amp:     ADSRinit_dB(0, 6.978f, 127, 6.978f); break;
```
JohannesLorenz, line 331: Just to make clear for the reader that it's a float:
```suggestion
        case ad_voice_amp:     ADSRinit_dB(0.f, 6.978f, 127, 6.978f); break;
```

JohannesLorenz (changes requested):
Only 2 comments. Please also change the "no used" to "not used".

Invite to re-review when you're finished.

src/Params/EnvelopeParams.cpp
friedolino78: you are right.

JohannesLorenz (approved):
All good. You can merge it after CI has passed.

PR #97, opened 2020-12-05
Topic/smooth filter cutoff and volume

original-male:
Use a LPF smoothing filter for cutoff frequency changes and apply in 8 sample chunks.

This removes a zipper artifact that was apparent when using keyboard aftertouch
to modulate filter cutoff.  This commit introduces generic
Value_Smoothing_Filter class (taken from Non-DAW), which can be employed
anywhere else in the application that such smoothing may be required.

A smoothing filter cutoff frequency of 10hz was found by ear to be the minimum
required to suppress the zipper effect.

The filter only runs when the cutoff value is being changed, so the
steady-state performance characteristics are unchanged.

 Master: Use Value_Smoothing_Filter for part and master volume smoothin too (incudes part pan).

### DIFF
--- a/src/DSP/AnalogFilter.cpp
+++ b/src/DSP/AnalogFilter.cpp
@@ -46,6 +46,7 @@ AnalogFilter::AnalogFilter(unsigned char Ftype,
     firsttime  = true;
     coeff.d[0] = 0; //this is not used
     outgain    = 1.0f;
+    freq_smoothing.sample_rate(samplerate_f);
 }
 
 AnalogFilter::~AnalogFilter()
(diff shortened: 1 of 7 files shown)


fundamental:
Excellent. Thanks for the addition. I'll try to play around with the new feature this week and I'm guessing the unit tests can be fixed by overriding the smoothing in the case of testing scenarios.

original-male:
I didn't run into any test errors when building this locally. Is there some extra level of testing I should have enabled?

fundamental:
On 12-07, Jonathan Moore Liles wrote:
> I didn't run into any test errors when building this locally.
> Is there some extra level of testing I should have enabled?

Yep. So, 'make test' will run through the unit tests, but only the ones
the system could build. For zyn the main thing of interest is setting up
CxxTest. It used to be pretty easy when it was initially adopted, but
it's a bit of a pain now. Looking on the PR you should see CI execution
which has the failed tests.

The zyn unit tests mostly verify that the same parameters result in the
same audio which this addition is currently failing (though it should be
a semi-easy fix).

PR #98, opened 2020-12-09
Topic/smooth filter cutoff and volume

original-male:
Same as before, but now don't fade up from zero initially, in order to hopefully pass CI tests.

### DIFF
--- a/src/DSP/AnalogFilter.cpp
+++ b/src/DSP/AnalogFilter.cpp
@@ -20,6 +20,10 @@
 #include "../Misc/Util.h"
 #include "AnalogFilter.h"
 
+
+const float MAX_FREQ = 20000.0f;
+const float MAX_FREQ_CO = 1.0f / MAX_FREQ;
+
 namespace zyn {
 
 AnalogFilter::AnalogFilter(unsigned char Ftype,
@@ -32,7 +36,8 @@ AnalogFilter::AnalogFilter(unsigned char Ftype,
       stages(Fstages),
       freq(Ffreq),
       q(Fq),
-      gain(1.0)
+     gain(1.0),
+     recompute(true)
 {
     for(int i = 0; i < 3; ++i)
         coeff.c[i] = coeff.d[i] = oldCoeff.c[i] = oldCoeff.d[i] = 0.0f;
@@ -43,7 +48,7 @@ AnalogFilter::AnalogFilter(unsigned char Ftype,
     coeff.d[0] = 0; //this is not used
     outgain    = 1.0f;
     freq_smoothing.sample_rate(samplerate_f);
-    freq_smoothing.reset( freq );
+    freq_smoothing.reset( freq * MAX_FREQ_CO );
 }
 
 AnalogFilter::~AnalogFilter()
@@ -257,7 +262,7 @@ AnalogFilter::Coeff AnalogFilter::computeCoeff(int type, float cutoff, float q,
     return coeff;
 }
 
-void AnalogFilter::computefiltercoefs(void)
+void AnalogFilter::computefiltercoefs(float freq, float q)
 {
     coeff = AnalogFilter::computeCoeff(type, freq, q, stages, gain,
             samplerate_f, order);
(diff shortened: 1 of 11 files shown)


original-male:
I tried a few other things and I was able to get all the tests except for PADNoteTest, but only by rasiing the smoothing filter cutoff far beyond the point were the zipper noise returned.

IMHO, the tests are just expecting the old lower quality output and there's nothing to do but change the tests.

original-male:
Oh, and one thing I forgot to mention: You can also use this for declicking if you set the volume to zero while you perform the click generating operation (such as switching programs) and then set it back to what it should be when done.

original-male:
Did the SVFilter too. Cleaned up a few things. Only the Formant filter remains using the old interpolation method.

original-male:
Now the formant filter is done too. However, upon further testing the SVFilter appears to have the zipper effect, I'm not sure what the problem is there, but I reverted it that one for the time being.

fundamental:
Curious. The state variable form should be the one which sounds the best in a time-varying sense, but I guess a fair amount of effort has been put into the direct form-II version of the other filter to help things sound cleaner.

fundamental:
Looks like the SVFilter code was implemented reasonably enough. There's the option of still doing the approach with par_interpolation, but I guess that sounds reasonable enough as is. Well done @dsheeler :)

Looking at the test cases it does look like there's envelopes triggering the new code, so yeah different output is to be expected. Output sounds quite nice, no audible zipper issues.

Out of curiosity is the cast on Value_Smoothing_Filter.cpp:46 needed? It looks like it's just trying to cast away the restrict specifier, but if that's the case it seems to defeat the purpose of it in the first place.

All that's really left is updating the tests with the new values (basically looking at every failure and copying over the new value into the unit tests), and some formatting issues (whitespace at the end of lines or on lines otherwise devoid of content and tabs used instead of spaces).
Let me know if a) there's other tweaks incoming and if you prefer to handle the formatting/unit tests or if you'd rather someone on zyn's side handle them.

original-male:
@fundamental Just did a little more testing. The old/current SVFilter does exhibit the zipper effect for me, but much less than AnalogFilter did before. I still haven't dug into why converting it to using the value smoothing filter didn't improve the situation (actually made it worse). I believe the only possible reason could be something happening in SVFilter::computefiltercoefs(). I think it would be best to convert SVFilter if possible before updating the unit test. The only other thing still using the old interpolation method is ADnote, which employs it in several places. I'm not sure what the sonic impacts of doing that are. It may be possible to get some smoother sound out of ADnote by replacing those parts. 

Regarding the cast in Value_Smoothing_Filter, that was just me working around the absence in the ZynAddSubFX codebase of some macros used in Non-DAW to do with gcc pramas involving buffer alignment. In Non-DAW everything is pretty well benchmarked and tuned to the max (to give an example, on my AMD hardware while (nframes--) is about twice as fast as for ( int i = 0; i < nframes; ++i )) IIRC I saw some significant speedups with SSE code by ensuring the alignment. Maybe GCC is smarter now though and doesn't require the hints. Anyway, if you want I can rewrite the code to look more sensible to a human, or include the macro/pragma magic it was trying to use.

Regarding the formatting, is there an emacs mode line or other standardized why to apply the formatting? I'm happy to do it as long as I don't have to manually cound spaces.

original-male:
@fundamental belay that. I think I just figured it out. Typo on my part. Patch for SVFilter forthcoming.

original-male:
OK. SVFilter now exhibits no zipper that I can hear. FYI for testing this stuff using keyboard aftertouch and a bright instrument (brass works well) and high pitched notes makes it most noticeable. It tends to be worse at the low frequency end of the filtter cuttoff than at the high end. If playing a sine wave bass type instrument I doubt you'd ever notice it.

original-male:
I tried my best with doing the formatting manually. Maybe one day I'll get a contribution to one my projects and I can complain about fomatting too ;-)

original-male:
@fundamental I've pushed another commit to improve the performance of AnalogFilter (probably faster than it was before this PR) and enforce some sanity limits. I noticed in testing/debugging a lot of filter cutoff values well in excess of 20kHz (coming from LFOs or envelopes, I have no idea), which IMHO can be safely ignored. I also noticed many fractional Hz changes to the cutoff frequency, which IMHO aren't audible and quantizing to 1Hz increments sounds fine to me. Any opinion on this?

PR #99, opened 2020-12-13
FilterParams: Fix timestamps updating on request

JohannesLorenz:
This came up by a compiler warning. Should be easy to review.


JohannesLorenz:
Actually, this change is obvious. IMO there's no review required (I just wanted to check CI).

Merging now.

PR #100, opened 2020-12-15
Sndio: Fix always true conditionals

JohannesLorenz:


fundamental:
Interesting that the man page isn't more explicit about the return code. The interpretation appears to be right based upon me reading between the lines, though I didn't look at the examples, just the man page.

JohannesLorenz:
Let's merge this then? Or should we make tests?

kinichiro:
sio_write and mio_read return size_t value and it never could be negative.
I think this PR is fine.

JohannesLorenz:
I just did a few tests with sndio, and it all worked. Time to merge this.

PR #101, opened 2020-12-17
Use an LPF smoothing filter for filter cutoff frequency changes, and …

original-male:
…master, and part volume/pan changes.

This removes a zipper artifact that was apparent when using keyboard aftertouch
to modulate filter cutoff.  This commit introduces generic
Value_Smoothing_Filter class (taken from Non-DAW), which can be employed
anywhere else in the application that such smoothing may be required.

A smoothing filter cutoff frequency of 10Hz was found by ear to be the minimum
required to suppress the zipper effect.

The filter only runs when the cutoff value is being changed, so the
steady-state performance characteristics are unchanged.

Same as #98, but squashed.

### DIFF
--- a/src/DSP/AnalogFilter.cpp
+++ b/src/DSP/AnalogFilter.cpp
@@ -32,20 +32,18 @@ AnalogFilter::AnalogFilter(unsigned char Ftype,
       stages(Fstages),
       freq(Ffreq),
       q(Fq),
-      gain(1.0),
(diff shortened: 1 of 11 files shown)


fundamental:
Uh. I guess this is bad timing, but I was mentioning on IRC earlier today that I was going to go through the open PRs tonight, make any needed modifications, and merge this evening.

PR #102, opened 2020-12-17
Add a .desktop to run zynaddsubfx without options, mandatory to make MIDI work

tobiasBora:
When you run `zynaddsubfx` without any options, it seems to be equivalent to:

    zynaddsubfx -I alsa -O jack

which means that it takes the input from alsa, and sends the output to Jack. It turns out that for some reason (no idea why), any midi device is recognize as an alsa device, at least using qjackctl:

image

so if you run `zynaddsubfx -I jack -O jack` instead, then you will get a midi device that is impossible to connect to any other midi device:

image

Therefore, I don't really see why people would want to run `zynaddsubfx -I jack -O jack`... But it's what people will run if they naively click on the existing .desktop entry `ZynAddSubFX - Jack`.

This PR tries to mitigate that issue by making sure that people would, by default, click a version that can indeed connect to MIDI external keyboards.

### DIFF
--- a/zynaddsubfx.desktop
+++ b/zynaddsubfx.desktop
@@ -1,0 +1,11 @@
+[Desktop Entry]
+Name=ZynAddSubFX
+Comment=A powerful realtime software synthesizer
+Comment[fr]=Un synthétiseur logiciel temps-réel puissant
+Comment[pl]=Funkcjonalny syntezator wirtualny czasu rzeczywistego
+Keywords=audio;sound;jack;midi;synth;synthesizer;
+Exec=zynaddsubfx
+Icon=zynaddsubfx
+Terminal=false
+Type=Application
+Categories=AudioVideo;Audio;


original-male:
Just to chime in on this issue: This affects the Non-Session-Manager experience as well. I (NSM author) am one of those strange people who wants to run with -I jack and -O jack at all times. You may be unaware of the -X seq option to jackd (check it out). It is much more convenient than dealing with alsa sequencer (well it used to be before the JACK devs anonymized and randomly enumerated all the system MIDI ports).

What I have proposed is that there be symlinks and argv[0] inspection so that e.g. "zynaddsubfx-jack" could be used by NSM users like myself, and "zynaddsubfx-alsa" by the oldschool people who still avoid JACK. Regular "zynaddsubfx" would presumably remain what you get today, which is apparently whatever you've compilied in support for. Accounting for all the permutations this way is cumbersome.

IMHO, desktop files are too distant from the underlying cause (and it doesn't help with NSM unless you've got it setup to allow external launchers to add clients to the current session).

falkTX:
I think this PR is misguided.
Mixing ALSA and JACK is not very optimal, because then the device wont be available to other applications. In a typical scenario for JACK, you want modularity, not applications hogging the devices.
Stuff like -X seq jackd option exists, also a2jmidid, and more recently pipewire (which will basically replace jackd in most generic distros soon) has its own version of a alsa-midi-jack bridge just like a2jmidid.

These many options are confusing, I think the issue is really somewhere else.
preferably zyn would provide an option in its gui that allows the user to setup the in/out as prefered. then have the CLI shortcuts/aliases/links that start up a preconfigured combo.

Carla works exactly this way, the "main" application started via desktop file follows user preferences.
But there is a carla-rack, carla-patchbay and carla-jack-multi CLI tools to start with exactly that mode regardless of user settings

tobiasBora:
@original-male @falkTX First, thanks for pointing out the `-X seq` option, it's interesting. Just, how am I supposed to translate it to QJackctl? Is it this box to uncheck like that? (sadly, I cannot test if it works because I don't have an external keyboard with me now)
image

However, I'm not sure of understanding your point @original-male, do you mean that NSM users would be penalized if user decide to use the .desktop I propose?

So for me the point of this PR is just to provide decent defaults for the newbie users. I'm not claiming that `-I alsa` is more efficient or even preferred (I agree with the fact that it seems better to handle MIDI directly from Jack instead of MIDI). I'm claiming that lot's of people interested by `zynaddsubfx` are people that are not jack expert and just want to be able to plug the keyboard/rosegarden/..., and be able to play some sounds using `zynaddsubfx` (it was me few years ago, or a friend recently). I'm not concerned about Jack experts, these people will already know the tools you mention (`-X seq`, a2midid...). And experts may not even use the `.desktop` files, they may even have there own, custom, optimized command.

So now, what happens when you are new to jack and want to use an external keyboard and/or rosegarden:
1) you realize that you need to use `zynaddsubfx` to play sounds
2) you realize that you need to use jack to play `zynaddsubfx` sounds
3) you realize that qjackctl is the recommended tool to use with jack and start it
4) you realize that pulseaudio applications does not work anymore. You lose 2 days to fix it.
5) you try to play notes by clicking on the keyboard of `zynaddsubfx`: it's not playing sound when you click :-(
6) you realize after few hours that you need to go in "Graph" in qjackctl and connect the keyboard to the system output
7) you plug your external keyboard: it's not playing sound when press a key. And you can't connect them in QJackctl.
8) you cry (and go back on Windows :-P).

So to make that a bit more simple (just removing steps 7 & 8), I see 3 solutions:
1) changing QJackctl to run by default the `-X seq`
2) creating a new `zynaddsubfx` .desktop, that new comers will likely click first and that uses `-I alsa` by default
3) writing a big warning in `zynaddsubfx` to explain precisely what you need to do to use MIDI input.

Since solution 2 is the simplest to implement... Here we are.

Concerning pipewire, it looks quite interesting. Bug guess we won't see that happenning really soon.

falkTX:
that would be the option to enable in qjackctl yes.
the issue with NSM is that CLI args are not allowed, and since it targets session management and modularity, the default of jack-midi and audio-midi makes sense for it.

But seeing from what you wrote, I think the main issue here is the use of standalones to achieve things that plugins already do quite well.
I understand you like rosegarden, but if it just implemented either LV2 and VST2, we could have zynaddsubfx working there without having to deal with any of this mess.
For plugins, the host sends audio and midi data directly, no need to deal with jack or alsa or connecting anything. Zyn has a DSSI version which can be loaded in rosegarden, but it only supports loading presets (there is no custom GUI due to DSSI design limitations not being able to save custom state)

In my opinion, users should really not have to deal with all this, understanding alsa vs jack-midi, options on a sound server etc.
Just gets in the way and makes users want to leave and do other things..

Are there any plans in Rosegarden to extend its plugin support? Maybe we would do best to focus on attention there.
Zyn could have yet another desktop file sure, but it would make it more confusing for the next new user that will then have 3 options to start zyn.

tobiasBora:
> that would be the option to enable in qjackctl yes.
Ok thanks.

> But seeing from what you wrote, I think the main issue here is the use of standalones to achieve things that plugins already do quite well.

What do you mean by plugins? Is it something that a user need to be aware of, install or something? If yes, then I think it would still add to the confusion, for me plugins are for stuff that are nice to have but not mandatory in a program. And playing sound... sounds like it's mandatory :-P It's like buying a car and when you get the car, you realize that the wheels are not included.

I never heard of LV2/VST2, but I guess it won't solve the issue for external keyboards right? They will still be recognized as ALSA devices.

> In my opinion, users should really not have to deal with all this, understanding alsa vs jack-midi, options on a sound server etc.
Just gets in the way and makes users want to leave and do other things..

Completely agree.

I'd say it's less confusing to have 3 desktop files with one that works rather than 2 with none that work. People will naturally try the 3 until it works. Like my friend told me "I don't understand, why, there is an Alsa version and a Jack version. In the Alsa version, I can type on the keyboard but I don't hear sound, in the Jack version I can hear sound but not type on the keyboard".

After, I agree that it would be quite nice to be able to change the input MIDI mode directly from within the program, at least people could try it directly from there instead of stopping/restarting the app (ideally with a bit of an explanation to guide the user). But I don't know zynaddsubfx good enough to be able to do this change myself.

falkTX:
> > 
> 
> What do you mean by plugins? Is it something that a user need to be aware of, install or something? If yes, then I think it would still add to the confusion, for me plugins are for stuff that are nice to have but not mandatory in a program. And playing sound... sounds like it's mandatory :-P It's like buying a car and when you get the car, you realize that the wheels are not included.
> 
> I never heard of LV2/VST2, but I guess it won't solve the issue for external keyboards right? They will still be recognized as ALSA devices.

Plugins are basically extensions to a host application. Instead of having 2 separate tools - rosegarden and zynaddsubfx - you would only have rosegarden and then load zynaddsubfx *inside* of it.

The tricky part with standalones (basically what you are using now) is that even after everything works, you will then still have to face the issue of saving that session so you can continue working on it later on.
How exactly do you plan on doing this once zynaddsubfx works? Manually reopening the correct files in each application takes time and is error prone. plus if you need to backup and reuse in another PC, you need to ensure you save all files in the same dir and/or backup everything together.
NSM tries to solve this by basically doing a session manager, where you simply "add" applications to a session, and it takes care to save them all in the same place and with a single button (Pressing "save" on NSM GUI would propagate the save signal to applications within the session, so everything gets saved at once)

WIth plugins though, you dont need to deal with any of this. The plugin is loaded inside the host, as if it was a native generator/synth/fx whatever, and:
1- audio and midi is sent to the plugin without having to connect anything anywhere (except only if you have a mixer with buses and stuff)
2- audio and midi OUT is sent to the host as well, same thing regarding connections, no need for a patchbay to manage those
3- plugin state (in this case zynaddsubfx) is saved together with the host project. there is no need to manually load anything
4- plugins do not have to care about dealing with MIDI or audio interfaces, it becomes the hosts job. if the host can connect and use your MIDI keyboard and speakers, so does the plugin

So it would solve your issue indirectly because then it is only the host that needs to be able to talk to the MIDI hw.

tobiasBora:
I see, thanks a lot for the explanation. So I see how it could help in rosegarden. And if you have lots of softwares that you want to connect, I guess that you start to see the use of NSM since it's unlikely that you can find a host with one plugin for each of your applications.

Now, if your only goal is to play music from a keyboard on zynaddsubfx in realtime, I guess that here you want to connect zynaddsubfx to jack directly, there is no reason to add another software right?

falkTX:
sometimes there is a reason still, where a plugin host can be modular and do all the usual stuff (talk to audio and midi hw) but work in a way that is familiar to you.
I am speaking from experience since I created the "carla plugin host" with this in mind. instead of needing to add some menus/dialogs to every plugin to work as standalone (so we can pick which hardware to use, for example), it can all just be done in a simple host.
so when you got a plugin host like this setup, you can very easily add synths or similar plugins and have it "just work" because everything was set-up already before. the host does the heavy lifting, the plugins just need to process audio and midi.

most of recent software that does some kind of sound synthesis or sample playback is now made with this in mind, so they ship in a plugin format instead of a standalone executable/binary.

tobiasBora:
Ok I see thanks. I thought that each plugin was written for a specific program, but as I understand they seem to be more reusable.

original-male:
For the record, @falkTX is not the creator of NSM, he is not the maintainer of NSM, he is not an authority on NSM, he is not a contributor to NSM, he is not an expert on NSM. Therefore one should disregard his statements about what is and is not appropriate with regard to NSM. What is he then? Well, he's very skilled at bypassing github's Block User feature in order to follow me around and harass me.

Anyway, ignoring this troublesome interlocutor...

There are two acceptable solutions to this problem from NSM's perspective:

1) The aforementioned symlink/argv[0] trick
2) A client should have a user editable global default for new sessions/projects, and a user configurable setting within the session/project, that is saved and loaded with the session/project. This ensures that the user can change their mind at any time with regard to their preference for JACK or ALSA without breaking existing sessions, and it allows them to always invoke the client by one name/desktop link, and to never involve command line arguments.

My preference is for #2, but it's considerably more complicated to implement than #1 (although not very complicated in absolute terms).

Consider that a user may have some sessions where ZynAddSubFX users ALSA and some where it uses JACK, and may have done this knowingly, intentionally, and for valid reasons.

I'm guessing that #2 is most likely to make @tobiasBora content (even outside of the context of NSM). This seems like a relatively straight forward first contribution project. The trickiest part is figuring out what to do about all the existinig projects/sessions out there which leave the NIO settings up in the air. There's probably not much that can be done about that, and the global default would apply in that case just as if it were a new session.

fundamental:
So, just checking in this conversation. I'd say that the best thing would be the global user configurable input/output settings. IMO, not too terribly hard to add (they can be swapped at runtime and a global config is read/saved for zyn already), though the input/output engine swapping never got exposed in the newer fusion GUI, so that might be something which needs to be done.

JohannesLorenz:
There has been no interest in discussion in the last ~6 months, so I'm closing this. The discussions are also a bit off-topic from the PR content. If you guys want to keep discussing, open a separate issue.

Btw, if you want to make a desktop icon, e.g. in KDE you can simply drag-drop zyn from the startup menu onto your desktop and further configure it there (right click on the icon).

Thanks anyways for the submission.

PR #103, opened 2020-12-22
Fix compiler warning (unused variable)

JohannesLorenz:
Additional comments are added to explain that `--embed` runs zyn as a
plugin, and that window embedding only works with NTK.

falkTX, can you please review this?


JohannesLorenz:
@fundamental , I did a rebase/force push because the compiler warning was already silenced. What remains is the improvement of the help text.

I hope you haven't been busy 5 months trying to figure out what this PR does? :laughing:

fundamental:
Looks like I archived this one without reading closely enough. The comment changes look good to me.

PR #104, opened 2020-12-22
Fix compiler warnings

JohannesLorenz:
Almost all warnings for gcc 10.2.0 and clang 11.0.0 are gone after this PR (together with #100 and #103).

### DIFF
--- a/src/Misc/MiddleWare.cpp
+++ b/src/Misc/MiddleWare.cpp
@@ -1096,7 +1096,7 @@ static int extractInt(const char *msg)
 
 static const char *chomp(const char *msg)
 {
-    while(*msg && *msg!='/') ++msg; \
+    while(*msg && *msg!='/') ++msg;
     msg = *msg ? msg+1 : msg;
     return msg;
 };
--- a/src/Synth/OscilGen.cpp
+++ b/src/Synth/OscilGen.cpp
@@ -200,9 +200,6 @@ const rtosc::Ports OscilGen::non_realtime_ports = {
             d.broadcast("/damage", "s", repath);
         }}};
 
-//unused:
-//#define rForwardCb [](const char *msg, rtosc::RtData &d) {\
-//    printf("forwarding...\n"); d.forward();}
 const rtosc::Ports OscilGen::realtime_ports{
     rSelf(OscilGen),
     rPresetType,
--- a/src/Synth/OscilGen.cpp
+++ b/src/Synth/OscilGen.cpp
@@ -269,7 +269,8 @@ const rtosc::MergePorts OscilGen::ports{
 //operations on FFTfreqs
 inline void clearAll(fft_t *freqs, int oscilsize)
 {
-    memset(freqs, 0, oscilsize / 2 * sizeof(fft_t));
+    fft_t zero = 0;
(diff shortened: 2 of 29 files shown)


JohannesLorenz:
Compiler error detected in DPF (submodule updated), submitted as DISTRHO/DPF#247 .

src/UI/FormantFilterGraph.cpp
```diff
@@ -63,7 +63,7 @@ void FormantFilterGraph::OSC_value(int x, const char *loc)
 }
 void FormantFilterGraph::OSC_value(unsigned x, void *v)
 {
-    assert(x = sizeof(Pvowels));
+    assert(x == sizeof(Pvowels));
```
fundamental, line 66: Oh dear. That was a silly mistake of (likely) mine.

src/Misc/MiddleWare.cpp
```diff
@@ -1096,7 +1096,7 @@ static int extractInt(const char *msg)
 
 static const char *chomp(const char *msg)
 {
-    while(*msg && *msg!='/') ++msg; \
+    while(*msg && *msg!='/') ++msg;
```
fundamental, line 1099: Ah, that must have come from the old SNIP macro.

src/Synth/OscilGen.cpp
```diff
@@ -269,7 +269,8 @@ const rtosc::MergePorts OscilGen::ports{
 //operations on FFTfreqs
 inline void clearAll(fft_t *freqs, int oscilsize)
 {
-    memset(freqs, 0, oscilsize / 2 * sizeof(fft_t));
+    fft_t zero = 0;
+    std::fill_n(freqs, oscilsize / 2, zero);
```
fundamental, line 273: I'd hope this turns into a memset under the hood. C++ standard should guarantee that std::complex<float> is the same as "_complex float" in C and that representation should be identical in memory as float[2]. Either way, it's better to fix the warning than fret about a few cycles difference.

fundamental:
Seems good. Feel free to merge unless there's a desire to wait for the aforementioned DPF fix. (since tests are passing I'm guessing you selected some intermediary commit in that submodule.

src/Synth/OscilGen.cpp
JohannesLorenz: I just found an interesting answer about exactly that on [StackOverflow](https://stackoverflow.com/a/51658588). It says we can either use `static_assert(std::is_trivially_copyable<std::complex<float>>::value);` before `memset`, or use `std::fill`, which is opitimized on GCC and clang, but not on MSVC. What do you think after reading this? I'm OK with either solution.

JohannesLorenz:
Strange, it seems like just my CMake cache was outdated. I removed the CMake dir and rebuilt, and now it all works. Let me know about the `std::fill`, then I'll merge.

src/Synth/OscilGen.cpp
fundamental: Well, as long as GCC/Clang are happy. AFAIK MSVC still hates the codebase due to parts of the rtosc port structure, so it's not really a supported dev platform anyhow.

PR #105, opened 2020-12-28
Topic/disentangle volume and expression

original-male:
This fixes some bad behavior regarding the interaction between volume and expression controls and seems to fix another insanely-loud-output bug.

### DIFF
--- a/src/DSP/AnalogFilter.cpp
+++ b/src/DSP/AnalogFilter.cpp
@@ -274,22 +274,22 @@ void AnalogFilter::setfreq(float frequency)
     if(frequency < 0.1f)
         frequency = 0.1f;
     else if ( frequency > MAX_FREQ )
-	frequency = MAX_FREQ;
+        frequency = MAX_FREQ;
 
     float rap = freq / frequency;
     if(rap < 1.0f)
         rap = 1.0f / rap;
 
     frequency = ceilf(frequency);/* fractional Hz changes are not
-				 * likely to be audible and waste CPU,
-				 * esp since we're already smoothing
-				 * changes, so round it */
+                                 * likely to be audible and waste CPU,
+                                 * esp since we're already smoothing
+                                 * changes, so round it */
 
     if ( fabsf( frequency - freq ) >= 1.0f )
     {
-	/* only perform computation if absolutely necessary */
-	freq = frequency;
-	recompute = true;
+        /* only perform computation if absolutely necessary */
+        freq = frequency;
+        recompute = true;
     }
 }
 
@@ -356,8 +356,8 @@ void AnalogFilter::singlefilterout(float *smp, fstage &hist)
 
     if ( recompute )
     {
-	computefiltercoefs(freq,q);
-	recompute = false;
+        computefiltercoefs(freq,q);
+        recompute = false;
     }
 
     if(order == 1) {  //First order filter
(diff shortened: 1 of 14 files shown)


src/DSP/AnalogFilter.cpp
```diff
@@ -383,24 +388,74 @@ void AnalogFilter::singlefilterout(float *smp, fstage &hist,
...
+
+    if ( freq_smoothing.apply( freqbuf, buffersize, freq * MAX_FREQ_CO ) )
+    {
+	/* in transition, need to do fine grained interpolation */
+	for(int i = 0; i < stages + 1; ++i)
+	    singlefilterout_freqbuf(smp, history[i], freqbuf);
+    }
+    else
+    {
+	/* stable state, just use one coeff */
```
JohannesLorenz, line 456: Did you use 2 spaces for indentation (also in many other places in this PR)? It should be 4...

original-male: I don't trouble myself with such things, and get whatever EMacs gives me in terms of indentation. Going against the grain is rather tedious. Is there an EMacs mode line or indent script or something equivalent that I run the code through to automatically comply with the formatting requirements?

JohannesLorenz: We don't supply such a script. You can fix it manually, or "allow edits from maintainers" in the PR to let one of us do it manually.

fundamental: We do have some really old scripts. I don't know if they work any more though (it was back when the initial yoshimi fork happened). style.cfg and style.sh in the root directory which uses 'uncrustify'. I'm guessing "allow edits from maintainers" is the simpler option.

original-male: It appears to me that "allow edits by maintainers" is already checked by default.

original-male: And FYI, what you're looking at currently is already my best effort to "do it manually." Clearly that's not adequate.

original-male: So, what do you guys do personally? Do you count spaces manually?

fundamental: I'm not sure about everyone else, but with vim, it's mostly shiftwidth=4, expandtab, textwidth=80 for indentation.

original-male: It's been a while since I used vim for anything other than editing a configuration file. I'm assuming "shiftwidth=4" means a tab is 4 spaces and "expandtab" means the saved file will contain no tabs? If there are no tabs, then I guess that explains the confusion, because the EMacs default is to indent with tabs.

original-male:
I ran "M-x untabify" on the files I touched. Was that sufficient to pass the style constraints?

JohannesLorenz:
> I ran "M-x untabify" on the files I touched. Was that sufficient to pass the style constraints?

If you mean 8bd5d8e , then no (see in the "Files changed" tab).

original-male:
What exactly am I supposed to see? How are you guys measuring this stuff? Shouldn't this be a commit hook or something? This is getting absurd. Am I supposed to just keep guessing how to comply with the unstated style requirements?

JohannesLorenz:
See [these screenshots](https://imgur.com/a/HX2G3sx) from the "Files changed" tab.

We can also do this for you if it's too difficult for you to fix this.

fundamental:
Honestly I was going to make the modifications myself, there's no need for you to do them on your end, my schedule has just been dreadful over the past few weeks myself and I hadn't carved out a timeslot to do it myself.

PR #106, opened 2021-01-10
[CMAKE] Fix configuration error on Windows.

carlo-bramini:
I'm trying to compile ZynAddSubFx with MinGW-W64, but I got this error when CMake configured the build:

```
CMake Error at src/CMakeLists.txt:455 (string):
  string sub-command STRIP requires two arguments.
```

While CMakeLists.txt file does syntactically contain two arguments, ${FLTK_LDFLAGS} can be empty. If this happens, then the string() function never receives the correct number of parameters, which then sees just one. If the string could be empty, then CMakeLists.txt must quote it.

EDIT: I just discovered that into `doc/win64-mingw64-msys2/README.txt` there is a `sed` command for patching the CMake file, but actually it is a bit unknown to me the reason because it has not been applied directly to `CMakeLists.txt`, since in my opinion it should be the right way to do it.


fundamental:
Thanks for the addition. The msys doc was merged as provided IIRC, and it does sound like the modification should have been directly applied.

PR #107, opened 2021-01-11
Smooth filter cutoff and fix volume and expression interaction

original-male:
As before, rebased onto current master (with conflict resolved).

### DIFF
--- a/src/DSP/AnalogFilter.cpp
+++ b/src/DSP/AnalogFilter.cpp
@@ -274,22 +274,22 @@ void AnalogFilter::setfreq(float frequency)
     if(frequency < 0.1f)
         frequency = 0.1f;
     else if ( frequency > MAX_FREQ )
-	frequency = MAX_FREQ;
+        frequency = MAX_FREQ;
 
     float rap = freq / frequency;
     if(rap < 1.0f)
         rap = 1.0f / rap;
 
     frequency = ceilf(frequency);/* fractional Hz changes are not
-				 * likely to be audible and waste CPU,
-				 * esp since we're already smoothing
-				 * changes, so round it */
+                                 * likely to be audible and waste CPU,
+                                 * esp since we're already smoothing
+                                 * changes, so round it */
 
     if ( fabsf( frequency - freq ) >= 1.0f )
     {
-	/* only perform computation if absolutely necessary */
-	freq = frequency;
-	recompute = true;
+        /* only perform computation if absolutely necessary */
+        freq = frequency;
+        recompute = true;
     }
 }
 
@@ -356,8 +356,8 @@ void AnalogFilter::singlefilterout(float *smp, fstage &hist)
 
     if ( recompute )
     {
-	computefiltercoefs(freq,q);
-	recompute = false;
+        computefiltercoefs(freq,q);
+        recompute = false;
     }
 
     if(order == 1) {  //First order filter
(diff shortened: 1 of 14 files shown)


fundamental:
If the merge conflict is resolved, steps from my end will be:

1) get the code locally, run all unit tests and update them with new constants. The previous version of this PR was reviewed and the output sounded right as well as the code looked sound, so as long as there aren't modifications between then and now, it should be fine to use the new output values as the expected output.
2) Run through any of the areas changed and tidy up any minor formatting differences, though I imagine most of the changes are at least partially addressed in commit 0abeffe

fundamental:
Looks like one of the tests may have found a problem, investigating...

fundamental:
Looks like at least one of the asserts is incorrect as the dB range for part volume is -40 dB to +13.333 dB, so gains above 1 are built in (whether that's a good idea is something else). I'll check the other asserts.

original-male:
Isn't that in conflict with the MIDI inspired side of things where 127 (the max possible to encode) equals 0dB?

original-male:
Also, related and regarding the -40dB thing. I changed regarding the processing of CC 7, as my expectation as a user based on all other MIDI equipment I've used is that CC 7 == 0 means no output (i.e. muted). I don't think this specifically was casuing the problem with CC 7 and CC 11 interfering, but it seemed to me that it was adding to the confusion.

fundamental:
The -40dB..13.3333 dB should be a result of looking at what the prior 0..127 knob mapped onto. I'm guessing that's what you're referring to, but it's also possible you're talking about the additional expression modifier where your argument might hold true.  The test failure I'm seeing is in relation to the OSC API getting and fetching the 'Volume' parameter and getting different values. I'm not seeing modifications which would trigger that yet though.

fundamental:
> I changed the processing of CC 7 ...

Good to know, I'll make a mental note to add that change in behavior to the changelogs where user visible changes tend to get documented.

fundamental:
Ok, I think I found it. I'm guessing that the Int(0..127)->Float(-40...13.3333) conversion code has a bug which one of the unit tests is highlighting (good thing since the coverage with version-to-version upgrade code is somewhat sparse).

Fixing now...

original-male:
Expression behaved as expected before this PR (and after) where 0 is totally muted and 127 is 0dB.

It may be possible to take out the changes I made regarding the gain ranges, but I'm not 100% sure one or more of them isn't necessary to solve the bad interaction between volume and expression. I kind of took the shotgun approach of just fixing everything that looked fishy.

fundamental:
No problem, I'm guessing now that I've got a lead it will be resolved in about 7 minutes

original-male:
(and most of what looked fishy was confusion between "volume" and "Volume" hence the slight renaming of members and methods to make them more easily distinguishable.)

fundamental:
Yeah the volume/Volume thing was a bit annoying. The typical convention for old parameters was 'Pparametername' (0..127 int) which are mostly getting updated to 'parametername' (float with real units). Volume was missing that convention in the old code, hence uglyness in updating it.


Diff right now which fixes the last unit test which is loading from an older serialized file.

```
 float Part::volume127TodB(unsigned char volume_)
 {
     assert( volume_  <= 127 );
-
-    return 20.0f * log10f(volume_ / 127.0f);
+    return (volume_ - 96.0f) / 96.0f * 40.0;
 }
 
 void Part::setVolumedB(float Volume_)
@@ -954,14 +953,14 @@ void Part::setVolumedB(float Volume_)
 
     Volume_ = limit(Volume_, -40.0f, 13.333f);
 
-    assert(Volume_ < 40.0);
+    assert(Volume_ < 14.0);
     Volume = Volume_;
 
     float volume = dB2rap( Volume_ );
 
-    /* printf( "Volume: %f, Expression %f\n", volume, ctl.expression.relvolume ); */
+    printf( "Volume: %f, Expression %f\n", volume, ctl.expression.relvolume );
 
-    assert( volume <= 1.0f );
+    assert( volume <= dB2rap(14.0f) );
 
     gain = volume * ctl.expression.relvolume;
 }
@@ -1301,7 +1300,7 @@ void Part::getfromXML(XMLwrapper& xml)
     if (xml.hasparreal("volume")) {
         setVolumedB(xml.getparreal("volume", Volume));
     } else {
-        setVolumedB(volume127TodB( xml.getpar127("volume", gain * 127.0f )));
+        setVolumedB(volume127TodB( xml.getpar127("volume", 96)));
     }
     setPpanning(xml.getpar127("panning", Ppanning));
```

original-male:
BTW, were you able to reproduce the CC 7 / C 11 interaction with the old code? The way I found it was by using CC 7 to set effectively the upper limit of volume for using an expression pedal (with my foot moving it slightly to feel it out). Got blasted a couple of times and then put that printf in to see what was really going on. Expression value was always reasonable (due to being mapped rather directly from the MIDI), but the volume (gain multiplier) was all over the place, sometimes 100x. Would be good to retest that specifically if any changes are made to the dB calculation.

fundamental:
I was able to find the one place where the volume was getting loaded as something crazy like +50dB and that bug should now be resolved. I'm totally in favor leaving the asserts in place in case volume gets too loud. Better to crash than blast someone. Right now I'm just changing their upper bound to the one established by the allowable volume range rather than maxing out at +0dB.

original-male:
Regarding that last hunk, the part for getting the real value uses the current Volume (dB) value as the fallback, shouldn't the one for reading the 7-bit value do something similar?

fundamental:
It can, but IIRC with the new code the 0..127 mapped value isn't stored anymore. So the code would have to do float->127 conversion to establish what to do when the volume is missing. The only case I can think of is loading a zyn fork like yoshimi's files or something. Falling back to the default volume 96 -> +0dB seems like the lazier implementation option (to me at least).

original-male:
This was unrelated to that first loud volume bug, if that's what you're referring to. I was already running a version including your fix for that when I got blasted by this one (which I could only trigger by adjusting CC 7 and CC 11 at the same time).

fundamental:
Darn another loud volume bug? Yeah I don't think I had any fixes for that after correcting for the first one. If you think that's still in place let me know and I can throw in another limiter in the style of "gain = limit(gain, lower, upper)". It's the one in Part.cpp which was causing it, right?

fundamental:
Ok, unit tests updated, 1 assert fixed, old .xmz loading bug fixed, now to check other asserts and fix random indentation.

original-male:
Yes, the first was due to mixed up units in when loading the part. The one fixed by this PR was due (I believe) to mixed up units when updating the volume in order to apply the expression control gain factor. I haven't experienced any loud blasts since implementing this PR (but I definitely feel more comfortable with the assertions in there for protection).

fundamental:
Ok, so the first one I was able to have a fix by looking for when the synth tries to load things incorrectly (+50dB IIRC) and replaces it with +0.0dB. Then this PR has another large volume fix and adds some additional asserts (keep in mind that if you're in release mode I think the asserts are gone, I forget about RelWithDebInfo).

original-male:
True. Probably a good idea to clamp it too for the release build users.

fundamental:
The value = limit(value, min, max) code is still there, just not the additional assert(value <= max);

fundamental:
ok, hopefully travis-ci doesn't end up spitting out any failures (if it does I just have to loosen the tolerance on a check or two), but this should be all set now.

fundamental:
Does this PR combine all of the previous PRs that were open or is there additional code which might resurface again? (no constraints on when, just trying to keep track of that sort of stuff)

original-male:
This is all of it. Figured it would be better to have it all in one place before messing with the unit tests.

I did try to apply some similar optimizations to SVFilter that I did to AnalogFilter, but these reintroduced the zipper effect (never got to the bottom of ti even after spending quite some time on it), so that's not included and probably not forthcoming.

fundamental:
After you had mentioned the SVFilter stuff I spent of bit of time reading the code myself. I couldn't find anything obvious which would have resulted in the zippering that you experienced. Lots of newer options like the SV formulation over the typical Direct form II (I think) biquad, but zyn devs have hacked on the 'analogfilter' a lot more.

Thanks for the patch set. Hopeful in the future I manage to keep my queue/backlog better in line as I hate being on the other end of laggy contribution workflows.

original-male:
Excellent! Thanks.

PR #110, opened 2021-02-05
Implement ZynAddSubFX::loadProgram

JohannesLorenz:
This is a draft PR that I could not test, because I don't know how to use this functionality. falkTX said that these functions can not be accessed by Lv2, but by using jack "directly", but I gave up trying to find out how. fundamental said that I could copy the code mostly from `src/Nio/InMgr.cpp`.

As I could not test this, it's questionable if anything of this works. 

Intention of this PR: Fix compiler warnings (unused parameters).

### DIFF
--- a/src/Plugin/ZynAddSubFX/ZynAddSubFX.cpp
+++ b/src/Plugin/ZynAddSubFX/ZynAddSubFX.cpp
@@ -273,7 +273,7 @@ protected:
     */
     void initProgramName(uint32_t index, String& programName) override
     {
-        programName = "Default";
+        programName = middleware->getProgramName(index).c_str();
     }
 
    /**
@@ -282,7 +282,8 @@ protected:
     */
     void loadProgram(uint32_t index) override
     {
-        setState(nullptr, defaultState);
+        //in plugin mode, only the first part is useful
+        middleware->pendingSetProgram(0, index);
     }
 
    /* --------------------------------------------------------------------------------------------------------
--- a/src/Misc/MiddleWare.cpp
+++ b/src/Misc/MiddleWare.cpp
@@ -2433,6 +2433,11 @@ void MiddleWare::pendingSetProgram(int part, int program)
     impl->bToU->write("/setprogram", "cc", part, program);
 }
 
+std::string zyn::MiddleWare::getProgramName(int program) const
+{
+    return impl->master->bank.ins[program].name;
+}
+
 std::string MiddleWare::activeUrl(void) const
 {
     return impl->last_url;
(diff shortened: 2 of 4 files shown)


src/Plugin/ZynAddSubFX/ZynAddSubFX.cpp
```diff
@@ -282,7 +282,10 @@ class ZynAddSubFX : public Plugin
     */
     void loadProgram(uint32_t index) override
     {
-        setState(nullptr, defaultState);
+        for(int i=0; i < NUM_MIDI_PARTS; ++i) {
+            //set the program of the parts assigned to all midi channels
+            middleware->pendingSetProgram(i, index);
```
JohannesLorenz, line 287: `src/Output/DSSIaudiooutput.cpp` has similar code. However, in DPF, there seems to be no MIDI part(?), so I set the program to all parts. Probably not good?

fundamental: I'd stick to the assumption that the user expected behavior is to have only one instrument running in zyn when in plugin mode. So, load it once into the 0th program?

JohannesLorenz: You mean in plugin mode, there's only one part ("part 1") useful, so use only `middleware->pendingSetProgram(0, index);`?

fundamental: Technically the other parts are still there. If it wasn't more work to disable the extra parts it would likely make sense to do so in the plugin. Generally only the lowest index part is something I'd expect users to want to use.

JohannesLorenz:
Anyone here who can test this PR? Or at least anyone who can describe how exactly this PR can be tested, setting a program in Zyn via Jack, but without Lv2?

AnClark:
Seems great. Does it have effect on VST?

JohannesLorenz:
> Seems great. Does it have effect on VST?

@falkTX do you know the answer?

falkTX:
DPF does not expose the programs for the VST version, as doing so would be very costly (spec says we need to store in-memory all presets and their changes as part of the plugin state).
So no, this will not affect VST.

Note that it is also not used for LV2 in runtime either, as in LV2 the host is the one in charge of loading plugin state.
For LV2, the only time this is used is when generating the TTL, as DPF "ttl export" goes through the list of programs and dumps its info to disk.

JohannesLorenz:
I really don't know how to test this (and no one else does), but the code fixes compiler warnings about unused variables, and we should really get our PRs closed. @fundamental can we just merge this? If errors will be reported, then at least we have a tester.

fundamental:
Merging with the idea that if this causes issues we'll just revert this single commit.

PR #113, opened 2021-02-14
Fixed a few compiler warnings

pgervais:
Fixed some minor string buffer size issues, which could lead to non-zero terminated strings or buffer overflow.

Fixed an "unused variable" warning due to code being preprocessed out.


fundamental:
Looks good. Thanks for the addition. :+1:

PR #114, opened 2021-03-02
add LFO fade in feature

friedolino78:
this adds a small feature, that I saw in a tutorlas for creating pad sounds in Serum
It lets the LFO fadein after the start delay over a specified amount of time.
used for frequency the spectrum can look like this:
[https://imgur.com/a/G3d8j9H](https://imgur.com/a/G3d8j9H)

### DIFF
--- a/src/Misc/Time.h
+++ b/src/Misc/Time.h
@@ -54,6 +54,11 @@ class RelTime
         {
             return (lhs.frame-rhs.frame);
         }
+        
+        friend bool operator>(const RelTime& lhs, const RelTime& rhs)
+        {
+            return (lhs.frame>rhs.frame);
+        }
     private:
         int64_t frame;
         int32_t sample;
--- a/src/Misc/Time.h
+++ b/src/Misc/Time.h
@@ -48,17 +48,6 @@ class RelTime
         bool inThisFrame() {return t.time() == frame;};
         bool inPast() {return t.time() > frame;}
         bool inFuture() {return t.time() < frame;}
-        int64_t over() {return t.time() - frame;}
-        
-        friend int64_t operator-(const RelTime& lhs, const RelTime& rhs)
-        {
-            return (lhs.frame-rhs.frame);
-        }
-        
-        friend bool operator>(const RelTime& lhs, const RelTime& rhs)
-        {
-            return (lhs.frame>rhs.frame);
-        }
     private:
         int64_t frame;
         int32_t sample;
--- a/src/Misc/Time.h
+++ b/src/Misc/Time.h
@@ -48,6 +48,12 @@ class RelTime
         bool inThisFrame() {return t.time() == frame;};
         bool inPast() {return t.time() > frame;}
         bool inFuture() {return t.time() < frame;}
+        int64_t over() {return t.time() - frame;}
+        
+        friend int64_t operator-(const RelTime& lhs, const RelTime& rhs)
+        {
+            return (lhs.frame-rhs.frame);
+        }
     private:
         int64_t frame;
         int32_t sample;
--- a/src/Params/LFOParams.cpp
+++ b/src/Params/LFOParams.cpp
@@ -81,8 +81,11 @@ static const rtosc::Ports _ports = {
               rPreset(ad_voice_amp, 0.94),
               "Delay before LFO start\n0..4 second delay"),
     rParamF(fadein, rShort("fadein"), rSpecial(disable), rUnit(S),
-              rLinear(0.0, 10.0), rDefault(0),
+              rLinear(0.0, 10.0f), rDefault(0.0f),
               "Time to ramp up LFO amplitude\n0..10 second"),
+    rParamF(fadeout, rShort("fadeout"), rSpecial(disable), rUnit(S),
+              rLinear(0.001, 10.0f), rDefault(10.0f),
+              "Time to ramp down LFO amplitude\n0..10 second"),
     {"Pdelay::i", rShort("delay") rLinear(0,127)
      rDoc("Delay before LFO start\n0..4 second delay"), NULL,
     [](const char *msg, RtData &d)
@@ -149,7 +152,7 @@ void LFOParams::setup()
 
 // TODO: reuse
 LFOParams::LFOParams(const AbsTime *time_) :
-    LFOParams(2.65, 0, 0, 127, 0, 0, 0, 0, 0, loc_unspecified, time_)
+    LFOParams(2.65, 0, 0, 127, 0, 0, 0.0, 0.0, 0.0, 0, loc_unspecified, time_)
 {
 }
 
@@ -161,6 +164,7 @@ LFOParams::LFOParams(float freq_,
                      char Prandomness_,
                      float Pdelay_,
                      float Pfadein_,
+                     float Pfadeout_,
                      char Pcontinous_,
                      consumer_location_t loc,
                      const AbsTime *time_) : loc(loc),
@@ -174,6 +178,7 @@ LFOParams::LFOParams(float freq_,
     Drandomness = Prandomness_;
     Ddelay      = Pdelay_;
     Dfadein     = Pfadein_;
+    Dfadeout    = Pfadeout_;
     Dcontinous  = Pcontinous_;
 
     setup();
@@ -186,7 +191,7 @@ LFOParams::LFOParams(consumer_location_t loc,
 
     auto init =
         [&](float freq_, char Pintensity_, char Pstartphase_, char Pcutoff_, char PLFOtype_,
-            char Prandomness_, float delay_, float fadein_, char Pcontinous_)
+            char Prandomness_, float delay_, char Pcontinous_)
     {
         Dfreq       = freq_;
         Dintensity  = Pintensity_;
@@ -195,18 +200,19 @@ LFOParams::LFOParams(consumer_location_t loc,
         DLFOtype    = PLFOtype_;
         Drandomness = Prandomness_;
         Ddelay      = delay_;
-        Dfadein     = fadein_;
+        Dfadein     = 0.0f;
+        Dfadeout    = 10.0f;
         Dcontinous  = Pcontinous_;
     };
 
     switch(loc)
     {
-        case ad_global_amp:    init(6.49, 0, 64, 127, 0, 0, 0, 0, 0); break;
-        case ad_global_freq:   init(3.71, 0, 64, 127, 0, 0, 0, 0, 0); break;
-        case ad_global_filter: init(6.49, 0, 64, 127, 0, 0, 0, 0, 0); break;
-        case ad_voice_amp:     init(11.25, 32, 64, 127, 0, 0, 0.94, 0, 0); break;
-        case ad_voice_freq:    init(1.19, 40,  0, 127, 0, 0,  0, 0, 0); break;
-        case ad_voice_filter:  init(1.19, 20, 64, 127, 0, 0,  0, 0, 0); break;
+        case ad_global_amp:    init(6.49, 0, 64, 127, 0, 0, 0, 0); break;
+        case ad_global_freq:   init(3.71, 0, 64, 127, 0, 0, 0, 0); break;
+        case ad_global_filter: init(6.49, 0, 64, 127, 0, 0, 0, 0); break;
+        case ad_voice_amp:     init(11.25, 32, 64, 127, 0, 0, 0, 0); break;
+        case ad_voice_freq:    init(1.19, 40,  0, 127, 0, 0, 0, 0); break;
+        case ad_voice_filter:  init(1.19, 20, 64, 127, 0, 0, 0, 0); break;
         default: throw std::logic_error("Invalid LFO consumer location");
     }
 
@@ -226,6 +232,7 @@ void LFOParams::defaults()
     Prandomness = Drandomness;
     delay       = Ddelay;
     fadein      = Dfadein;
+    fadeout     = Dfadeout;
     Pcontinous  = Dcontinous;
     Pfreqrand   = 0;
     Pstretch    = 64;
@@ -269,6 +276,9 @@ void LFOParams::getfromXML(XMLwrapper& xml)
     if (xml.hasparreal("fadein")) {
         fadein      = xml.getparreal("fadein", fadein);
     }
+    if (xml.hasparreal("fadeout")) {
+        fadeout      = xml.getparreal("fadeout", fadeout);
+    }
     Pstretch    = xml.getpar127("stretch", Pstretch);
     Pcontinous  = xml.getparbool("continous", Pcontinous);
 }
@@ -285,6 +295,7 @@ void LFOParams::paste(LFOParams &x)
     COPY(Pfreqrand);
     COPY(delay);
     COPY(fadein);
+    COPY(fadeout);
     COPY(Pcontinous);
     COPY(Pstretch);
 
--- a/src/Params/LFOParams.cpp
+++ b/src/Params/LFOParams.cpp
@@ -80,6 +80,9 @@ static const rtosc::Ports _ports = {
               rLinear(0.0, 4.0), rDefaultDepends(loc), rDefault(0),
               rPreset(ad_voice_amp, 0.94),
               "Delay before LFO start\n0..4 second delay"),
+    rParamF(fadein, rShort("fadein"), rSpecial(disable), rUnit(S),
+              rLinear(0.0, 10.0), rDefault(0),
+              "Time to ramp up LFO amplitude\n0..10 second"),
     {"Pdelay::i", rShort("delay") rLinear(0,127)
      rDoc("Delay before LFO start\n0..4 second delay"), NULL,
     [](const char *msg, RtData &d)
@@ -91,6 +94,7 @@ static const rtosc::Ports _ports = {
              obj->delay = 4.0f * rtosc_argument(msg, 0).i / 127.0f;
          }
      }},
+     
     rToggle(Pcontinous, rShort("c"), rDefault(false),
             "Enable for global operation"),
     rParamZyn(Pstretch, rShort("str"), rCentered, rDefault(64),
@@ -145,7 +149,7 @@ void LFOParams::setup()
 
 // TODO: reuse
 LFOParams::LFOParams(const AbsTime *time_) :
-    LFOParams(2.65, 0, 0, 127, 0, 0, 0, 0, loc_unspecified, time_)
+    LFOParams(2.65, 0, 0, 127, 0, 0, 0, 0, 0, loc_unspecified, time_)
 {
 }
 
@@ -156,6 +160,7 @@ LFOParams::LFOParams(float freq_,
                      char PLFOtype_,
                      char Prandomness_,
                      float Pdelay_,
+                     float Pfadein_,
                      char Pcontinous_,
                      consumer_location_t loc,
                      const AbsTime *time_) : loc(loc),
@@ -168,6 +173,7 @@ LFOParams::LFOParams(float freq_,
     DLFOtype    = PLFOtype_;
     Drandomness = Prandomness_;
     Ddelay      = Pdelay_;
+    Dfadein     = Pfadein_;
     Dcontinous  = Pcontinous_;
 
     setup();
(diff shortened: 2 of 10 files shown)


friedolino78:
corresponding zest PR:
https://github.com/mruby-zest/mruby-zest/pull/54

src/Params/LFOParams.h
```diff
@@ -64,6 +65,7 @@ class LFOParams:public Presets
         unsigned char Prandomness; /**<randomness (0=off)*/
         unsigned char Pfreqrand; /**<frequency randomness (0=off)*/
         float         delay; /**<delay (0=off)*/
+        float         fadein; /**<fadein (0=off)*/
```
JohannesLorenz, line 68:
```suggestion
        float         fadein; /**<fadein, relative to delay (0=off)*/
```

JohannesLorenz (changes requested):

src/Params/LFOParams.cpp
```diff
@@ -189,17 +195,18 @@ LFOParams::LFOParams(consumer_location_t loc,
...
 
     switch(loc)
     {
-        case ad_global_amp:    init(6.49, 0, 64, 127, 0, 0, 0, 0); break;
-        case ad_global_freq:   init(3.71, 0, 64, 127, 0, 0, 0, 0); break;
-        case ad_global_filter: init(6.49, 0, 64, 127, 0, 0, 0, 0); break;
-        case ad_voice_amp:     init(11.25, 32, 64, 127, 0, 0, 0.94, 0); break;
-        case ad_voice_freq:    init(1.19, 40,  0, 127, 0, 0,  0, 0); break;
-        case ad_voice_filter:  init(1.19, 20, 64, 127, 0, 0,  0, 0); break;
+        case ad_global_amp:    init(6.49, 0, 64, 127, 0, 0, 0, 0, 0); break;
```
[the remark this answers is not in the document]
friedolino78: yes. i had to do some manual merging locally for my all-in-branch.

src/Synth/LFO.cpp
```diff
@@ -193,7 +194,18 @@ float LFO::lfoout()
         out *= lfointensity * amp2;
 
     if(delayTime.inFuture())
+    {
+        outConst = out;
```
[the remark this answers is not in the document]
friedolino78: yes, i will fix it.

src/Params/LFOParams.cpp
```diff
@@ -80,6 +80,9 @@ static const rtosc::Ports _ports = {
               rLinear(0.0, 4.0), rDefaultDepends(loc), rDefault(0),
               rPreset(ad_voice_amp, 0.94),
               "Delay before LFO start\n0..4 second delay"),
+    rParamF(fadein, rShort("fadein"), rSpecial(disable), rUnit(S),
+              rLinear(0.0, 10.0), rDefault(0),
```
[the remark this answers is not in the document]
JohannesLorenz: Yes, default should be changed. I just meant if we add a new preset with a different default value. But the comment is not important, feel free to resolve it.

src/Tests/guitar-adnote.xmz
```diff
@@ -100,6 +100,7 @@ version-revision="6" ZynAddSubFX-author="Nasca Octavian Paul">
 <par name="randomness_amplitude" value="0" />
 <par name="randomness_frequency" value="0" />
 <par_real name="delay" value="0" exact_value="0x00000000" />
+<par_real name="fadein" value="0" exact_value="0x00000000" />
```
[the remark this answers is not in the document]
friedolino78: yes, it could load the preset without this param. but the load-save-compare test needs all things to be specified.

JohannesLorenz (commented):
Review passed. Only 2 minor comments.

Additional questions:

1. Do you want to add the knob for the FL UI?
2. Would there be a use case to abstract this feature into an Envelope? We already have delay and fade-in, I wonder if a full Envelope would be useful?

friedolino78:
> Review passed. Only 2 minor comments.
> 
> Additional questions:
> 
>     1. Do you want to add the knob for the FL UI?

I dont know if there is enough gui space left since I dont use it any more. but i would do it if there is enough demand.

> 
>     2. Would there be a use case to abstract this feature into an Envelope? We already have delay and fade-in, I wonder if a full Envelope would be useful?

envelope for lfo lets ring mod matrix i my ears.
while being possible it's not an as low hanging fruit as this one with a few lines of code.

src/Nio/InMgr.cpp
```diff
@@ -96,7 +96,7 @@ void InMgr::flush(unsigned frameStart, unsigned frameStop)
     MidiEvent ev;
     while(!work.trywait()) {
         queue.peak(ev);
-        if(ev.time < (int)frameStart || ev.time > (int)frameStop) {
+        if(ev.time < (long unsigned int)frameStart || ev.time > (long unsigned int)frameStop) {
```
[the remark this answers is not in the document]
friedolino78: actually i included it because of a similar warning. have to investigate this a bit further.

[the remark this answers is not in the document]
friedolino78: i removed the InMgr changes. it has nothing to do with fadein

src/Params/LFOParams.cpp
```diff
@@ -218,6 +225,7 @@ void LFOParams::defaults()
     PLFOtype    = DLFOtype;
     Prandomness = Drandomness;
     delay       = Ddelay;
+    fadein       = Dfadein;
```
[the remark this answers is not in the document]
friedolino78: yap.

friedolino78:
new zest PR using new submodule structure: 
https://github.com/mruby-zest/mruby-zest-build/pull/48

JohannesLorenz:
So, the status of this PR is:

1. The PR is OK by now.
2. You can add FL code, or not (you can still open a further PR for this). Shall I wait?

If we would ever implement an envelope for LFO, do you think it would be compatible with savefiles from this branch? I think it should be easy to convert fade-in time into an LFO?

friedolino78:
> So, the status of this PR is:
> 
> 1. The PR is OK by now.
> 2. You can add FL code, or not (you can still open a further PR for this). Shall I wait?

don't wait.

> 
> If we would ever implement an envelope for LFO, do you think it would be compatible with savefiles from this branch? I think it should be easy to convert fade-in time into an LFO?

it would be even possible to replace both delay and fadein time by the ENV and map it to the first two elements of envdt while loading. i also think it would be easy.

friedolino78:
> What do you think about adding the same for Release? Then it would be similar to the Attack/Release knobs in the Freq Envelope, and would probably cover 90% of uses of "we need an envelope for this".

that would be a fade out time. but when should that start? at noteoff?

JohannesLorenz:
> that would be a fade out time. but when should that start? at noteoff?

I think that's the only point of time that makes sense (and other envelopes do it like this, too).

friedolino78:
> > that would be a fade out time. but when should that start? at noteoff?
> 
> I think that's the only point of time that makes sense (and other envelopes do it like this, too).

ok. i already started implementing. but passing the noteoff to lfo is not as trivial as the fade in was.

src/Params/LFOParams.cpp
```diff
@@ -81,8 +81,11 @@ static const rtosc::Ports _ports = {
               rPreset(ad_voice_amp, 0.94),
               "Delay before LFO start\n0..4 second delay"),
     rParamF(fadein, rShort("fadein"), rSpecial(disable), rUnit(S),
-              rLinear(0.0, 10.0), rDefault(0),
+              rLinear(0.0, 10.0f), rDefault(0.0f),
               "Time to ramp up LFO amplitude\n0..10 second"),
+    rParamF(fadeout, rShort("fadeout"), rSpecial(disable), rUnit(S),
+              rLinear(0.001, 10.0f), rDefault(10.0f),
+              "Time to ramp down LFO amplitude\n0..10 second"),
```
JohannesLorenz, line 88: Typo: "seconds" (also in the fadein port). Maybe you should add ", 10 = none"?

src/Params/LFOParams.cpp
```diff
@@ -195,18 +200,19 @@ LFOParams::LFOParams(consumer_location_t loc,
...
-        case ad_global_amp:    init(6.49, 0, 64, 127, 0, 0, 0, 0, 0); break;
-        case ad_global_freq:   init(3.71, 0, 64, 127, 0, 0, 0, 0, 0); break;
-        case ad_global_filter: init(6.49, 0, 64, 127, 0, 0, 0, 0, 0); break;
-        case ad_voice_amp:     init(11.25, 32, 64, 127, 0, 0, 0.94, 0, 0); break;
-        case ad_voice_freq:    init(1.19, 40,  0, 127, 0, 0,  0, 0, 0); break;
-        case ad_voice_filter:  init(1.19, 20, 64, 127, 0, 0,  0, 0, 0); break;
+        case ad_global_amp:    init(6.49, 0, 64, 127, 0, 0, 0, 0); break;
+        case ad_global_freq:   init(3.71, 0, 64, 127, 0, 0, 0, 0); break;
+        case ad_global_filter: init(6.49, 0, 64, 127, 0, 0, 0, 0); break;
+        case ad_voice_amp:     init(11.25, 32, 64, 127, 0, 0, 0, 0); break;
```
JohannesLorenz, line 213: Did you change `0.94` to `0` here intentionally? I guess this was a mistake?

src/Params/LFOParams.cpp
```diff
@@ -269,6 +276,9 @@ void LFOParams::getfromXML(XMLwrapper& xml)
     if (xml.hasparreal("fadein")) {
         fadein      = xml.getparreal("fadein", fadein);
     }
+    if (xml.hasparreal("fadeout")) {
```
JohannesLorenz, line 279: `fadeout` is missing in `add2XML`?

src/Params/LFOParams.h
```diff
@@ -66,6 +67,7 @@ class LFOParams:public Presets
         unsigned char Pfreqrand; /**<frequency randomness (0=off)*/
         float         delay; /**<delay (0=off)*/
         float         fadein; /**<fadein, relative to delay (0=off)*/
+        float         fadeout; /**<fadein (0=off)*/
```
JohannesLorenz, line 70: Copy-Paste-Error for the comment.

src/Synth/LFO.cpp
```diff
@@ -28,8 +28,9 @@ LFO::LFO(const LFOParams &lfopars, float basefreq, const AbsTime &t, WatchManage
     fadeInTime(t, lfopars.delay+lfopars.fadein), //0..8 sec
     waveShape(lfopars.PLFOtype),
     deterministic(!lfopars.Pfreqrand),
-    dt_(t.dt()),
-    lfopars_(lfopars), basefreq_(basefreq),
+        dt_(t.dt()),
+    lfopars_(lfopars), 
+    basefreq_(basefreq),
```
JohannesLorenz, line 33: These 2/3 lines should not be changed at all, right? They are not related to this PR.

src/Tests/guitar-adnote.xmz
```diff
@@ -134,6 +135,7 @@ version-revision="6" ZynAddSubFX-author="Nasca Octavian Paul">
 <par name="randomness_amplitude" value="0" />
 <par name="randomness_frequency" value="0" />
 <par_real name="delay" value="0.598425" exact_value="0x3F193265" />
+<par_real name="fadein" value="0" exact_value="0x00000000" />
```
JohannesLorenz, line 138: Maybe, after you have solved the other comments, you will need to add `fadeout` here aswell?

src/Synth/ADnote.cpp
```diff
@@ -1936,6 +1936,9 @@ void ADnote::releasekey()
     NoteGlobalPar.FreqEnvelope->releasekey();
     NoteGlobalPar.FilterEnvelope->releasekey();
     NoteGlobalPar.AmpEnvelope->releasekey();
+    NoteGlobalPar.FreqLfo->releasekey();
+    NoteGlobalPar.FilterLfo->releasekey();
+    NoteGlobalPar.AmpLfo->releasekey();
```
JohannesLorenz, line 1941: I think this needs to be added for PAD and possibly for SUB. For the effects, we don't support fadein/fadeout, so it's just PAD and SUB?

JohannesLorenz (changes requested):

src/Params/LFOParams.cpp
friedolino78: yes

src/Params/LFOParams.cpp
friedolino78: ups

src/Synth/LFO.cpp
friedolino78: it's only code formatting for better readability.

src/Tests/guitar-adnote.xmz
friedolino78: ok

src/Synth/ADnote.cpp
friedolino78: SUB has no LFOs. I add it to PAD

src/Synth/LFO.cpp
```diff
@@ -154,6 +159,22 @@ float LFO::biquad(float input)
     return (cutoff==127) ? input : output; // at cutoff 127 bypass filter
 }
 
+void LFO::releasekey()
+{
+    if (lfopars_.fadeout==10.0f) // deactivated
+        return;
+    // store current ramp value in case of release while fading in
+    rampConst = ramp; 
+    outConst *= ((1.0f/ramp)-1.0f);
```
JohannesLorenz, line 168: I don't understand this statement (maybe comment it?). Also, it might end up with division by zero?

JohannesLorenz:
Btw, we have a segmentation fault in the watch test (with guitar-adnote.xmz).

src/Misc/Time.h
```diff
@@ -48,6 +48,17 @@ class RelTime
...
         bool inPast() {return t.time() > frame;}
         bool inFuture() {return t.time() < frame;}
+        int64_t over() {return t.time() - frame;}
+        
+        friend int64_t operator-(const RelTime& lhs, const RelTime& rhs)
+        {
+            return (lhs.frame-rhs.frame);
+        }
+        
+        friend bool operator>(const RelTime& lhs, const RelTime& rhs)
```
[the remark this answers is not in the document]
friedolino78: without friend it gives compiler errors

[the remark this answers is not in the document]
friedolino78: with const, too

src/Synth/LFO.cpp
```diff
@@ -193,7 +214,40 @@ float LFO::lfoout()
         out *= lfointensity * amp2;
 
     if(delayTime.inFuture())
+    {
         return out;
+    } else if (!fadein){
+        fadeInTime = lfopars_.time->time();
```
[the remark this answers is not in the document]
friedolino78: delayTime is a reltime object, fadeInTime is now a absloute timestamp to calculate the ramp up factor

src/Synth/LFO.cpp
friedolino78: > Btw, we have a segmentation fault in the watch test (with guitar-adnote.xmz).

i don't know how to debug them

friedolino78: > I don't understand this statement (maybe comment it?). Also, it might end up with division by zero?

i removed the division and commented it.

src/Misc/Time.h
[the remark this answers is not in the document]
friedolino78: actually i don't use these functions anymore. I removed them now.

src/Synth/LFO.cpp
friedolino78: segfaults fixed

JohannesLorenz:
Try the following:

1. go to ad synth's global tab
2. convert AMP env to free, and set its release point to +5 seconds, going to ~85 %
(i.e. the note will still sound 5 seconds after release with almost 100%)
3. in the LFO part (still AMP): depth->90, fadein 3.4, fadeout -> 8.7
4. tip the note shortly. then the note suddenly loses volume when you release the key

Adding your fix from f70d02d also into the `case delaying` did not fix it either.

src/Synth/LFO.h
```diff
@@ -18,8 +18,18 @@
 #include "../Misc/Time.h"
 #include "WatchPoint.h"
 
+
+
 namespace zyn {
 
+typedef enum lfo_state_type{
```
JohannesLorenz, line 25: Consider putting this as `private` into the class.

JohannesLorenz (commented):
Only minor comments left. Code looks mostly good.

One thing I notice is that you take almost 10 seconds fade out, it seems to be finished a bit before 10 seconds (at 7 or 8). Is that wanted?

src/Synth/LFO.cpp
```diff
@@ -154,46 +159,130 @@ float LFO::biquad(float input)
...
+    if (lfopars_.fadeout==10.0f) { // deactivated
+        fadeOutDuration = 0.0f;
+        return;
+    }
+    // store current ramp value in case of release while fading in
+    rampOnRelease = ramp; 
+    // burn in the current amount of outStartValue    
+    // therefor multiply its current reverse ramping factor 
+    // and divide by current ramp factor. It will be multiplied during fading out.
+    outStartValue *= ((ramp>0.001) ? (1.0f - ramp)/ramp : 1000);
```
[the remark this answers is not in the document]
JohannesLorenz: outdated comment - resolving.

src/Synth/LFO.cpp
```diff
@@ -154,46 +159,130 @@ float LFO::biquad(float input)
...
+        case fadingOut:
+            if(fadeOutDuration && ramp) {// no division by zero, please
+                ramp = rampOnRelease * (1.0f - (float)(lfopars_.time->time() - releaseTimestamp) / (float)fadeOutDuration);
+                ramp *= ramp; // square
+            }
+            else // no ramp down
+                ramp = 0.0f; 
+            
+            out += outStartValue;
+            out *= ramp;
```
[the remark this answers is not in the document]
JohannesLorenz: outdated comment - PLEASE IGNORE!

friedolino78:
> Only minor comments left. Code looks mostly good.
> 
> One thing I notice is that you take almost 10 seconds fade out, it seems to be finished a bit before 10 seconds (at 7 or 8). Is that wanted?

could be due to the squared ramp slope (parabolic) probably the values near zero are quite small.
but with linear slope the ending was too abrupt in my ear.

JohannesLorenz:
> could be due to the squared ramp slope (parabolic) probably the values near zero are quite small.
> but with linear slope the ending was too abrupt in my ear.

I also think the squared ramp is the reason. It's probably best to keep it that way. I'd say finish the minor comments and we can merge this :smiley:

src/Synth/LFO.h
friedolino78: ok

src/Synth/LFO.cpp
```diff
@@ -154,46 +159,130 @@ float LFO::biquad(float input)
...
+    // Apply intensity
     if(waveShape == LFO_SINE || waveShape == LFO_TRIANGLE)
         out *= lfointensity * (amp1 + phase * (amp2 - amp1));
     else
         out *= lfointensity * amp2;
+    
+    
+    // handle lfo state (delay, fade in, fade out)
+    switch(lfo_state) {
+        case firstTick:
```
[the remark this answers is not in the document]
friedolino78: ok

src/Synth/LFO.cpp
```diff
@@ -154,46 +159,130 @@ float LFO::biquad(float input)
...
-                //x -= 0.25f; //chance the starting phase
+                lfointensity = powf(2, lfopars_.Pintensity / 127.0f * 11.0f) - 1.0f; // [0...2047] cent
                 break;
         }
     }
-
-    float out = baseOut(waveShape, phase);
-
+    
+    out = baseOut(waveShape, phase);
```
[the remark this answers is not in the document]
friedolino78: i removed prepareFadeIn as you suggested in another comment.

JohannesLorenz (approved):
Rework and code OK.
Testing OK.

PR #115, opened 2021-03-10
keep ENV knobs in sync in free mode

friedolino78:
this has the advantage that the ADSR ports can be still used in freemode. use case especially ADSR controls via midi CC.

### DIFF
--- a/src/Params/EnvelopeParams.cpp
+++ b/src/Params/EnvelopeParams.cpp
@@ -45,7 +45,6 @@ namespace zyn {
             rLIMIT(var, atoi) \
             rCAPPLY(obj->name, "f", obj->name = dTREAL(var)) \
             data.broadcast(loc, "i", dT127(obj->name));\
-            rChangeCb \
         } rBOIL_END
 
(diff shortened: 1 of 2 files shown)


fundamental:
This PR is still marked as a draft/WIP. Is that still the case?

friedolino78:
> This PR is still marked as a draft/WIP. Is that still the case?

no, just forgot to switch it back

friedolino78:
> Code looks right, but to avoid a possible crash:
> Is it possible for Penvpoints-1 or Penvsustain-1 to become negative values at any point? I know in non-free mode you should be safe, but in free mode you can delete a lot of points which might make it possible for one of those member variables to be zero. If it's safe, let me know and this is good to merge

it's safe. 

Penvpoints is only decremented in line 291. But if it's <=3 the code cannot be reached.
because of line 283:
`        if(curpoint<1 || curpoint>=env->Penvpoints-1 || env->Penvpoints<=3)
            return;`

fundamental:
Thanks for the confirmation

PR #116, opened 2021-03-19
add step sequencer to addsynth

friedolino78:
for modulating amp, freq and filter

started it just for fun after seeing videos of korg wavestate. Maybe it can be usable for others.

### DIFF
--- a/src/Params/ADnoteParameters.cpp
+++ b/src/Params/ADnoteParameters.cpp
@@ -18,6 +18,7 @@
 #include "ADnoteParameters.h"
 #include "EnvelopeParams.h"
 #include "LFOParams.h"
+#include "SEQParams.h"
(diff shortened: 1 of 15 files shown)


friedolino78:
branch doesnt work. works only in combination with commits from my BpmSync branch

src/Params/ADnoteParameters.cpp
```diff
@@ -1024,7 +1086,11 @@ void ADnoteGlobalParam::getfromXML(XMLwrapper& xml)
         FreqLfo->getfromXML(xml);
         xml.exitbranch();
 
+        xml.enterbranch("FREQUENCY_SEQ");
```
fundamental, line 1089: Load failure happens here. you're unconditionally trying to enter a branch which may or may not exist in the source savefile.

PR #117, opened 2021-03-23
Bpm Sync internal

friedolino78:
this branch makes LFO frequency and some Effect delay times syncable to global time.tempo (with a rational factor multiplyer)
this allows relative frequencies and polyrhythmic LFO combinations in different voices or amp/freq/filter.

### DIFF
--- a/src/Effects/EffectMgr.cpp
+++ b/src/Effects/EffectMgr.cpp
@@ -28,10 +28,13 @@
 #include "Phaser.h"
 #include "../Misc/XMLwrapper.h"
 #include "../Misc/Util.h"
+#include "../Misc/Time.h"
 #include "../Params/FilterParams.h"
 #include "../Misc/Allocator.h"
 
 namespace zyn {
+
+static const float S2DELAY = 127.0f / 40.0f;
 
 #define rObject EffectMgr
 #define rSubtype(name) \
@@ -117,6 +120,100 @@ static const rtosc::Ports local_ports = {
                 }
             }
         }},
+    {"numerator::i", rShort("num") rDefault(0) rLinear(0,99)
+        rProp(parameter) rDoc("Numerator of ratio to bpm"), NULL,
+        [](const char *msg, rtosc::RtData &d)
+        {
+            EffectMgr *eff = (EffectMgr*)d.obj;
+            if(rtosc_narguments(msg)) {
+                int val = rtosc_argument(msg, 0).i;
+                if (val>=0) {
+                    eff->numerator = val;
+                    int Pdelay, Pfreq;
+                    float freq;
+                    if(eff->denominator) {
+                        switch(eff->nefx) {
+                        case 2: // Echo
+                            // invert:
+                            // delay = (Pdelay / 127.0f * 1.5f); //0 .. 1.5 sec
+                            Pdelay = (int)roundf((20320.0f / (float)eff->time->tempo) * 
+                                                 ((float)eff->numerator / (float)eff->denominator));
+                            if (eff->numerator&&eff->denominator)
+                                eff->seteffectparrt(2, Pdelay);
+                            break;
+                        case 3: // Chorus
+                        case 4: // Phaser
+                        case 5: // Alienwah
+                        case 8: // DynamicFilter
+                            freq =  ((float)eff->time->tempo * 
+                                     (float)eff->denominator / 
+                                     (240.0f * (float)eff->numerator));
+                            // invert:
+                            // (powf(2.0f, Pfreq / 127.0f * 10.0f) - 1.0f) * 0.03f
+                            Pfreq = (int)roundf(logf((freq/0.03f)+1.0f)/LOG_2 * 12.7f);
+                            if (eff->numerator&&eff->denominator)
+                                eff->seteffectparrt(2, Pfreq);
+                            break;
+                        case 1: // Reverb
+                        case 6: // Distorsion
+                        case 7: // EQ
+                        default:
+                            break;
+                        }
+                    }
+                }
+                d.broadcast(d.loc, "i", val);
+            } else {
+                d.reply(d.loc, "i", eff->numerator); 
+            }
+        }},
+    {"denominator::i", rShort("dem") rDefault(4) rLinear(1,99)
+        rProp(parameter) rDoc("Denominator of ratio to bpm"), NULL,
+        [](const char *msg, rtosc::RtData &d)
+        {
+            EffectMgr *eff = (EffectMgr*)d.obj;
+            if(rtosc_narguments(msg)) {
+                int val = rtosc_argument(msg, 0).i;
+                if (val > 0) {
+                    eff->denominator = val;
+                    int Pdelay, Pfreq;
+                    float freq;
+                    if(eff->numerator) {
+                        switch(eff->nefx) {
+                        case 2: // Echo
+                            // invert:
+                            // delay = (Pdelay / 127.0f * 1.5f); //0 .. 1.5 sec
+                            Pdelay = (int)roundf((20320.0f / (float)eff->time->tempo) * 
+                                                 ((float)eff->numerator / (float)eff->denominator));
+                            if (eff->numerator&&eff->denominator)
+                                eff->seteffectparrt(2, Pdelay);
+                            break;
+                        case 3: // Chorus
+                        case 4: // Phaser
+                        case 5: // Alienwah
+                        case 8: // DynamicFilter
+                            freq =  ((float)eff->time->tempo * 
+                                     (float)eff->denominator / 
+                                     (240.0f * (float)eff->numerator));
+                            // invert:
+                            // (powf(2.0f, Pfreq / 127.0f * 10.0f) - 1.0f) * 0.03f
+                            Pfreq = (int)roundf(logf((freq/0.03f)+1.0f)/LOG_2 * 12.7f);
+                            if (eff->numerator&&eff->denominator)
+                                eff->seteffectparrt(2, Pfreq);
+                            break;
+                        case 1: // Reverb
+                        case 6: // Distorsion
+                        case 7: // EQ
+                        default:
+                            break;
+                        }
+                    }
+                }
+                d.broadcast(d.loc, "i", val);
+            } else {
+                d.reply(d.loc, "i", eff->denominator); 
+            }
+        }},
     {"eq-coeffs:", rProp(internal) rDoc("Get equalizer Coefficients"), NULL,
         [](const char *, rtosc::RtData &d)
         {
@@ -173,6 +270,8 @@ EffectMgr::EffectMgr(Allocator &alloc, const SYNTH_T &synth_,
       nefx(0),
       efx(NULL),
       time(time_),
+      numerator(0),
+      denominator(4),
       dryonly(false),
       memory(alloc),
       synth(synth_)
@@ -243,6 +342,41 @@ void EffectMgr::changeeffectrt(int _nefx, bool avoidSmash)
                 efx = NULL;
                 break; //no effect (thru)
         }
+        
+        // set freq / delay params according to bpm ratio 
+        int Pdelay, Pfreq;
+        float freq;
+        if (numerator>0) {
+            switch(nefx) {
+                case 2: // Echo
+                    // invert:
+                    // delay = (Pdelay / 127.0f * 1.5f); //0 .. 1.5 sec
+                    Pdelay = (int)roundf((20320.0f / (float)time->tempo) * 
+                                         ((float)numerator / (float)denominator));
+                    if (numerator&&denominator)
+                        seteffectparrt(2, Pdelay);
+                    break;
+                case 3: // Chorus
+                case 4: // Phaser
+                case 5: // Alienwah
+                case 8: // DynamicFilter
+                    freq =  ((float)time->tempo * 
+                             (float)denominator / 
+                             (240.0f * (float)numerator));
+                    // invert:
+                    // (powf(2.0f, Pfreq / 127.0f * 10.0f) - 1.0f) * 0.03f
+                    Pfreq = (int)roundf(logf((freq/0.03f)+1.0f)/LOG_2 * 12.7f);
+                    if (numerator&&denominator)
+                        seteffectparrt(2, Pfreq);
+                    break;
+                case 1: // Reverb
+                case 6: // Distorsion
+                case 7: // EQ
+                default:
+                    break;
+            }
+        }
+        
     } catch (std::bad_alloc &ba) {
         std::cerr << "failed to change effect " << _nefx << ": " << ba.what() << std::endl;
         return;
@@ -488,6 +622,8 @@ void EffectMgr::add2XML(XMLwrapper& xml)
         xml.endbranch();
     }
     xml.endbranch();
+    xml.addpar("numerator", numerator);
+    xml.addpar("denominator", denominator);
 }
 
 void EffectMgr::getfromXML(XMLwrapper& xml)
@@ -533,6 +669,8 @@ void EffectMgr::getfromXML(XMLwrapper& xml)
         }
         xml.exitbranch();
     }
+    numerator = xml.getpar("numerator", numerator, 0, 99);
+    denominator = xml.getpar("denominator", denominator, 1, 99);
     cleanup();
 }
 
--- a/src/Effects/EffectMgr.h
+++ b/src/Effects/EffectMgr.h
@@ -73,6 +73,10 @@ class EffectMgr:public Presets
         int     nefx;
         Effect *efx;
         const AbsTime *time;
+        
+        int numerator;
+        int denominator;
+        
     private:
 
         //Parameters Prior to initialization
@@ -106,6 +110,8 @@ class EffectMgr:public Presets
         bool dryonly;
         Allocator &memory;
         const SYNTH_T &synth;
+        
+        
 };
 
 }
--- a/src/Params/LFOParams.cpp
+++ b/src/Params/LFOParams.cpp
@@ -30,6 +30,7 @@ namespace zyn {
 #define rChangeCb if (obj->time) { obj->last_update_timestamp = obj->time->time(); }
 #define rBegin [](const char *msg, rtosc::RtData &d) {
 #define rEnd }
+
 static const rtosc::Ports _ports = {
     rSelf(LFOParams),
     rPaste,
@@ -104,6 +105,10 @@ static const rtosc::Ports _ports = {
             "Enable for global operation"),
     rParamZyn(Pstretch, rShort("str"), rCentered, rDefault(64),
         "Note frequency stretch"),
+    rParamZyn(numerator, rShort("num"), rLinear(0,99), rDefault(0),
+        "Numerator of ratio to bpm"),
+    rParamZyn(denominator, rShort("dem"), rLinear(0,99), rDefault(4),
+        "Denominator of ratio to bpm"),
 // these are currently not yet implemented and must be hidden therefore
 #ifdef DEAD_PORTS
     //Float valued aliases
@@ -192,29 +197,29 @@ LFOParams::LFOParams(consumer_location_t loc,
                                              last_update_timestamp(0) {
 
     auto init =
-        [&](float freq_, char Pintensity_, char Pstartphase_, char Pcutoff_, char PLFOtype_,
-            char Prandomness_, float delay_, char Pcontinous_)
+        [&](float freq_, char Pintensity_, char Pstartphase_, float delay_)
     {
         Dfreq       = freq_;
         Dintensity  = Pintensity_;
         Dstartphase = Pstartphase_;
-        Dcutoff     = Pcutoff_;
-        DLFOtype    = PLFOtype_;
-        Drandomness = Prandomness_;
+        Dcutoff     = 127;
+        DLFOtype    = 0;
+        Drandomness = 0;
         Ddelay      = delay_;
         Dfadein     = 0.0f;
         Dfadeout    = 10.0f;
-        Dcontinous  = Pcontinous_;
+        Dcontinous  = 0;
+
     };
 
     switch(loc)
     {
-        case ad_global_amp:    init(6.49, 0, 64, 127, 0, 0, 0, 0); break;
-        case ad_global_freq:   init(3.71, 0, 64, 127, 0, 0, 0, 0); break;
-        case ad_global_filter: init(6.49, 0, 64, 127, 0, 0, 0, 0); break;
-        case ad_voice_amp:     init(11.25, 32, 64, 127, 0, 0, 0.94, 0); break;
-        case ad_voice_freq:    init(1.19, 40,  0, 127, 0, 0, 0, 0); break;
-        case ad_voice_filter:  init(1.19, 20, 64, 127, 0, 0, 0, 0); break;
+        case ad_global_amp:    init(6.49, 0, 64, 0.0f); break;
+        case ad_global_freq:   init(3.71, 0, 64, 0.0f); break;
+        case ad_global_filter: init(6.49, 0, 64, 0.0f); break;
+        case ad_voice_amp:     init(11.25, 32, 64, 0.94f); break;
+        case ad_voice_freq:    init(1.19, 40,  0, 0.0f); break;
+        case ad_voice_filter:  init(1.19, 20, 64, 0.0f); break;
         default: throw std::logic_error("Invalid LFO consumer location");
     }
 
@@ -238,6 +243,8 @@ void LFOParams::defaults()
     Pcontinous  = Dcontinous;
     Pfreqrand   = 0;
     Pstretch    = 64;
+    numerator   = 0;
+    denominator = 4;
 }
 
 
@@ -255,6 +262,8 @@ void LFOParams::add2XML(XMLwrapper& xml)
     xml.addparreal("fadeout", fadeout);
     xml.addpar("stretch", Pstretch);
     xml.addparbool("continous", Pcontinous);
+    xml.addpar("numerator", numerator);
+    xml.addpar("denominator", denominator);
 }
 
 void LFOParams::getfromXML(XMLwrapper& xml)
@@ -284,6 +293,9 @@ void LFOParams::getfromXML(XMLwrapper& xml)
     }
     Pstretch    = xml.getpar127("stretch", Pstretch);
     Pcontinous  = xml.getparbool("continous", Pcontinous);
+
+    numerator = xml.getpar("numerator", numerator, 0, 99);
+    denominator = xml.getpar("denominator", denominator, 0, 99);
 }
 
 #define COPY(y) this->y=x.y
@@ -300,6 +312,8 @@ void LFOParams::paste(LFOParams &x)
     COPY(fadein);
     COPY(fadeout);
     COPY(Pcontinous);
+    COPY(numerator);
+    COPY(denominator);
     COPY(Pstretch);
 
     if ( time ) {
--- a/src/Params/LFOParams.h
+++ b/src/Params/LFOParams.h
@@ -69,6 +69,8 @@ class LFOParams:public Presets
         float         fadein; /**<fadein, relative to delay*/
         float         fadeout; /**<fadeout on key release (10.0=off)*/
         unsigned char Pcontinous; /**<1 if LFO is continous*/
+        int           numerator;  /**<numerator for integer ratio between system tempo and LFO freq (0=off)*/
+        int           denominator;/**<denominator for integer ratio between system tempo and LFO freq (0=off)*/
         unsigned char Pstretch; /**<how the LFO is "stretched" according the note frequency (64=no stretch)*/
 
         //! what kind is the LFO (0 - frequency, 1 - amplitude, 2 - filter)
--- a/src/Synth/LFO.cpp
+++ b/src/Synth/LFO.cpp
@@ -21,42 +21,28 @@
 
 namespace zyn {
 
-LFO::LFO(const LFOParams &lfopars, float basefreq, const AbsTime &t, WatchManager *m,
+LFO::LFO(const LFOParams &lfopars_, float basefreq_, const AbsTime &t, WatchManager *m,
         const char *watch_prefix)
     :first_half(-1),
-    delayTime(t, lfopars.delay), //0..4 sec
-    waveShape(lfopars.PLFOtype),
-    deterministic(!lfopars.Pfreqrand),
-    dt_(t.dt()),
-    lfopars_(lfopars), 
-    basefreq_(basefreq),
+    time(t),
+    delayTime(t, lfopars_.delay), //0..4 sec
+    deterministic(!lfopars_.Pfreqrand),
+    dt(t.dt()),
+    lfopars(lfopars_), 
+    basefreq(basefreq_),
     watchOut(m, watch_prefix, "out")
 {
-    int stretch = lfopars.Pstretch;
-    if(stretch == 0)
-        stretch = 1;
-
-    //max 2x/octave
-    const float lfostretch = powf(basefreq / 440.0f, (stretch - 64.0f) / 63.0f);
-
-    const float lfofreq = lfopars.freq * lfostretch;
-    phaseInc = fabsf(lfofreq) * t.dt();
-
+    updatePars();
+    
     if(!lfopars.Pcontinous) {
-        if(lfopars.Pstartphase == 0)
+        if(!lfopars.Pstartphase)
             phase = RND;
         else
-            phase = fmod((lfopars.Pstartphase - 64.0f) / 127.0f + 1.0f, 1.0f);
+            phase = 0.0f;
     }
     else {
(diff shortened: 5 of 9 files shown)


friedolino78:
LFO phase is now rt capable. so you ca adjust the relative phase of fifferent synced LFOs while playing.

fundamental:
Looks like this branch isn't currently building under CI (and it doesn't look like an intermittent failure). I'm guessing that means that there's a needed rtosc update based upon it being an error within port sugar expansion.

friedolino78:
builds now, but test fails:

51/52 Test #51: PortChecker ......................***Failed    4.16 sec
sh: 1: /home/mk/synth/zynaddsubfx/build/src/Tests/../../rtosc/port-checker: not found
Error: port-checker has returned 127.

      Start 52: SaveOsc
52/52 Test #52: SaveOsc ..........................***Exception: Child aborted  0.56 sec
lo server running on 14191
Loading XML file...
Action /load_xmz finished after 0.058081 ms,
    with a timeout of <1 ms (success)
Saving OSC file now...
save-osc: /home/mk/synth/zynaddsubfx/rtosc/src/cpp/ports-runtime.cpp:203: size_t rtosc::helpers::get_value_from_runtime(void*, const rtosc::Port&, size_t, char*, const char*, char*, std::size_t, std::size_t, rtosc_arg_val_t*): Assertion `d.size() >= 0' failed.

fundamental:
I recall this getting discussed in IRC. Was there a resolution to the portchecker failure? If not let me know as I'd like to merge in this set in the near term.

JohannesLorenz:
Oh, I wanted to review this, but had no time yet :( . I think @friedolino78 solved the port checker issue locally (but not pushed yet), and `SaveOsc` is still failing.

fundamental:
Sounds good to me. I'll let you handle this one. Just ping me if one of the dependent issues requires a check from my end.

JohannesLorenz:
@friedolino78 It does not build anymore. See the travis-ci logs (I get the same compiler issues on my own PC).

friedolino78:
> @friedolino78 It does not build anymore. See the travis-ci logs (I get the same compiler issues on my own PC).

I have some changes stashed before creating the commit. Maybe that broke it. 
Will look after it later today.

JohannesLorenz:
SaveOsc is still failing. The reason is that the EffectMgr port "ratiofixed" claims to replay something with `d`, but the port's implementation does not reply anything:

```
save-osc: /home/johannes/cprogs/zyn/master/rtosc/src/cpp/ports-runtime.cpp:203: size_t rtosc::helpers::get_value_from_runtime(void*, const rtosc::Port&, size_t, char*, const char*, char*, std::size_t, std::size_t, rtosc_arg_val_t*): Assertion `d.size() >= 0' failed.

Thread 1 "save-osc" received signal SIGABRT, Aborted.
0x00007ffff71acef5 in raise () from /usr/lib/libc.so.6
(gdb) bt
#0  0x00007ffff71acef5 in raise () from /usr/lib/libc.so.6
#1  0x00007ffff7196862 in abort () from /usr/lib/libc.so.6
#2  0x00007ffff7196747 in __assert_fail_base.cold () from /usr/lib/libc.so.6
#3  0x00007ffff71a5646 in __assert_fail () from /usr/lib/libc.so.6
#4  0x00005555558fe757 in rtosc::helpers::get_value_from_runtime (runtime=0x5555565374e0, port=..., loc_size=8192, loc=0x7ffffffe5bc0 "/part0/partefx0/ratiofixed", portname_from_base=0x7ffffffed3a0 "ratiofixed", 
    buffer_with_port=0x7ffffffe7bc0 "ratiofixed", buffersize=8192, max_args=2048, arg_vals=0x7ffffffd9bc0) at rtosc/src/cpp/ports-runtime.cpp:203
```

Can you please try to fix this?

JohannesLorenz:
I didn't review completely, but there are a few issues:

1. Setting the tempo with hitting the plus and minus keys can make it require a lot of patience if you want to use 60 or 70...
2. If you set the LFO to e.g. 1/8 or 1/4 (default tempo 120), the LFO seems way to slow for my ears
3. At least the frequency LFO sounds broken. It does not make a correct sine. Instead, at the lower end, it seems to do a slope before it continues.

Considering 2 and 3, I [created a gist xmz](https://gist.github.com/JohannesLorenz/ba1656e2b59d15e83feccf9bba9b5b06) to explain the issue.

friedolino78:
> I didn't review completely, but there are a few issues:
> 
> 1. Setting the tempo with hitting the plus and minus keys can make it require a lot of patience if you want to use 60 or 70...
yes, that's not good.
I could imagine value dragging with the mouse like on knobs (easy to implement i guess) and double click to numerical input (could be harder).

> 2. If you set the LFO to e.g. 1/8 or 1/4 (default tempo 120), the LFO seems way to slow for my ears
could be the same problem as in 3. use a square reduce some depth and you will hear that at 1/8 the output changes every 2 seconds, which is what i would expect.

> 3. At least the frequency LFO sounds broken. It does not make a correct sine. Instead, at the lower end, it seems to do a slope before it continues.
> 
> Considering 2 and 3, I [created a gist xmz](https://gist.github.com/JohannesLorenz/ba1656e2b59d15e83feccf9bba9b5b06) to explain the issue.
What you are hearing is when the lfo makes the amp value crossing zero. Negative values sound like positive ones.
maybe amp value should be clipped at zero? but the phase change could be a intended feature if multiple voices are added.

JohannesLorenz:
OK, ignore point 1, I just realized I can use the mouse wheel to scroll through the values fast.

JohannesLorenz:
I would have expected that "1/8" changes the LFO duration to 1/8 note, which is 0.25 seconds at 120? Why do you expect 2 seconds?

friedolino78:
> I would have expected that "1/8" changes the LFO duration to 1/8 note, which is 0.25 seconds at 120? Why do you expect 2 seconds?

ah 1/8 is not a note but the relation to the system bpm value so 1/8 is 15bpm.

JohannesLorenz:
About point 3, if I use the xmz from the gist and turn the sync off (using the frequency knob old-style), I don't get the slope. Then it sounds like a plain sine. Which is really strange, because you did not change the code for LFO minimum and maximum...

JohannesLorenz:
It currently does not compile, because you renamed too many of the "denominator" in LFO.cpp.

friedolino78:
> It currently does not compile, because you renamed too many of the "denominator" in LFO.cpp.

fixed it. now even the test works :-)

src/Synth/LFO.cpp
```diff
@@ -154,38 +162,34 @@ float LFO::biquad(float input)
...
     }
+    
+    if (lfopars.numerator && lfopars.denominator && tempo != time.tempo) {
+        tempo = time.tempo;
+        float lfofreq = (float(tempo)) / 15.0f * lfopars.denominator/lfopars.numerator;
+        phaseInc = fabsf(lfofreq) * dt;
+    }
 
-    float out = baseOut(waveShape, phase);
+    float out = baseOut(waveShape, fmod(phase + (lfopars.Pstartphase + 63.0f) / 127.0f, 1.0f));
```
JohannesLorenz, line 192: Allowing the start phase to change while the LFO is running is actually a feature which is independent to BPM sync?

JohannesLorenz (changes requested):
Only few serious issues.

src/Effects/EffectMgr.cpp
```diff
@@ -117,6 +120,50 @@ static const rtosc::Ports local_ports = {
...
+    {"ratiofixed::i", rOptions(SPEEDRATIO_OPTIONS)
+        rProp(parameter) rLinear(0,21)
+        rDoc("select fixed ratio for BPM based delay") 
+        rDefault(off), NULL,
+        [](const char *msg, rtosc::RtData &d)
+        {
+            EffectMgr *eff = (EffectMgr*)d.obj;
+            if(rtosc_narguments(msg)) {
+                int val = rtosc_argument(msg, 0).i;
+                printf("rtosc_argument(msg, 0).i = %d\n", rtosc_argument(msg, 0).i);
```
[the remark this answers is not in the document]
friedolino78: ok

src/Synth/LFO.cpp
```diff
@@ -86,6 +73,28 @@ LFO::LFO(const LFOParams &lfopars, float basefreq, const AbsTime &t, WatchManage
...
+    const float lfostretch = powf(basefreq / 440.0f, (stretch - 64.0f) / 63.0f);
+
+    float lfofreq = lfopars.freq * lfostretch;
+    if (!lfopars.numerator || !lfopars.denumerator) {
+        lfofreq = lfopars.freq * lfostretch;   
+    } else {
+        tempo = time.tempo;
+        lfofreq = float(tempo) * float(lfopars.numerator)/(15.0f*float(lfopars.denumerator));
+    }
+    phaseInc = fabsf(lfofreq) * dt;
```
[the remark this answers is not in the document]
friedolino78: I see no problem. phaseInc is calculated before it is used.

src/Synth/LFO.cpp
```diff
@@ -21,42 +21,29 @@
...
+        if(!lfopars.Pstartphase)
             phase = RND;
         else
-            phase = fmod((lfopars.Pstartphase - 64.0f) / 127.0f + 1.0f, 1.0f);
+            phase = 0.0f;
     }
     else {
-        const float tmp = fmod(t.time() * phaseInc, 1.0f);
-        phase = fmod((lfopars.Pstartphase - 64.0f) / 127.0f + 1.0f + tmp, 1.0f);
+        phase = fmod((float)t.time() * phaseInc, 1.0f);
```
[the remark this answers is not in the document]
friedolino78: actually i dont know what line 60 is for.
but because of the fmod it doesnt hurt.

src/Synth/LFO.cpp
friedolino78: yes. but for synced LFOs it is more severely needed.

src/Effects/EffectMgr.cpp
```diff
@@ -117,6 +120,83 @@ static const rtosc::Ports local_ports = {
...
+                    case 1:
+                    case 6:
+                    case 7:
+                    default:
+                        break;
+                    }
+                int i;
+                int ind_last = ((int)sizeof(speedratios)/(int)sizeof(speedratios[0]));
+                for (i=0; i<ind_last; ++i) if (speedratios[i] == value) break;
+                d.reply(d.loc, "i", i);
```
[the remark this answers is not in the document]
friedolino78: i am switching to numerator/denominator now. speedratio will go.

src/Params/LFOParams.cpp
```diff
@@ -145,7 +187,7 @@ void LFOParams::setup()
 
 // TODO: reuse
 LFOParams::LFOParams(const AbsTime *time_) :
-    LFOParams(2.65, 0, 0, 127, 0, 0, 0, 0, loc_unspecified, time_)
+    LFOParams(2.65, 0, 0, 0, 0, 0, 0, 0, 0, loc_unspecified, time_)
```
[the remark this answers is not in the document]
friedolino78: could be a anrtifact from rebase, i had to manually merge a few things.

src/Params/LFOParams.h
```diff
@@ -65,6 +66,8 @@ class LFOParams:public Presets
         unsigned char Pfreqrand; /**<frequency randomness (0=off)*/
         float         delay; /**<delay (0=off)*/
         unsigned char Pcontinous; /**<1 if LFO is continous*/
+        unsigned char ratiofixed; /**<1 if LFO syncs to BPM*/
```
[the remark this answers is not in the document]
friedolino78: yes.

src/Params/LFOParams.cpp
```diff
@@ -179,27 +223,28 @@ LFOParams::LFOParams(consumer_location_t loc,
...
 
     switch(loc)
     {
-        case ad_global_amp:    init(6.49, 0, 64, 127, 0, 0, 0, 0); break;
-        case ad_global_freq:   init(3.71, 0, 64, 127, 0, 0, 0, 0); break;
-        case ad_global_filter: init(6.49, 0, 64, 127, 0, 0, 0, 0); break;
-        case ad_voice_amp:     init(11.25, 32, 64, 127, 0, 0, 0.94, 0); break;
-        case ad_voice_freq:    init(1.19, 40,  0, 127, 0, 0,  0, 0); break;
-        case ad_voice_filter:  init(1.19, 20, 64, 127, 0, 0,  0, 0); break;
+        case ad_global_amp:    init(6.49, 0, 64, 0, 0, 0, 0, 0); break;
```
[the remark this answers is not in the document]
friedolino78: i will cleanup there.

src/Params/LFOParams.cpp
```diff
@@ -179,27 +223,28 @@ LFOParams::LFOParams(consumer_location_t loc,
...
+        case ad_global_freq:   init(3.71, 0, 64, 0, 0, 0, 0, 0); break;
+        case ad_global_filter: init(6.49, 0, 64, 0, 0, 0, 0, 0); break;
+        case ad_voice_amp:     init(11.25, 32, 64, 0, 0.94, 0, 0, 0); break;
```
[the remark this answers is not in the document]
friedolino78: could be. the merge is a pita with that many zeros

src/Params/LFOParams.cpp
```diff
@@ -93,8 +94,10 @@ static const rtosc::Ports _ports = {
      }},
     rToggle(Pcontinous, rShort("c"), rDefault(false),
             "Enable for global operation"),
-    rParamZyn(Pstretch, rShort("str"), rCentered, rDefault(64),
-        "Note frequency stretch"),
+    rParamZyn(nominator, rShort("nom"), rDefault(0), rLinear(0,100),
```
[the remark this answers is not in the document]
friedolino78: ok

src/Effects/EffectMgr.cpp
```diff
@@ -117,6 +120,83 @@ static const rtosc::Ports local_ports = {
...
+                printf("speedratios[limit(val, 0, (int)sizeof(speedratios))] = %f\n", speedratios[limit(val, 0, (int)sizeof(speedratios))]);
+                printf("eff->time->tempo = %d\n", eff->time->tempo);
+                printf("S2DELAY = %f\n", S2DELAY);
+                
+                if (val) {
+                    int delay, Pfreq;
+                    float freq;
+                    switch(eff->nefx) {
+                    case 2:
+                        delay =  (int)(S2DELAY / eff->time->tempo * speedratios[limit(val, 0, (int)(sizeof(speedratios)/sizeof(speedratios[0]))-1)]);
```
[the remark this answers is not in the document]
friedolino78: no, just forgot about them

friedolino78:
now everything should work again.
Only issue: I would like to change is the denominator default, which imo should be 4.

src/Synth/LFO.cpp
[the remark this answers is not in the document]
JohannesLorenz: Sorry, I confused the lines. It's correct in your PR. Resolving my comment.

src/Params/LFOParams.h
```diff
@@ -69,6 +69,9 @@ class LFOParams:public Presets
         float         fadein; /**<fadein, relative to delay*/
         float         fadeout; /**<fadeout on key release (10.0=off)*/
         unsigned char Pcontinous; /**<1 if LFO is continous*/
+        float         speedratio; /**<ratio between system tempo and LFO freq (0=off)*/
```
JohannesLorenz, line 72: Still unused. Can you remove it, please?

src/Params/LFOParams.h
```diff
@@ -65,6 +65,10 @@ class LFOParams:public Presets
         unsigned char Pfreqrand; /**<frequency randomness (0=off)*/
         float         delay; /**<delay (0=off)*/
         unsigned char Pcontinous; /**<1 if LFO is continous*/
+        unsigned char ratiofixed; /**<1 if LFO syncs to BPM*/
+        float         speedratio;
```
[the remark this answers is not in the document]
JohannesLorenz: Replaced by newer comment, resolving.

src/Synth/SynthNote.cpp
```diff
@@ -184,6 +184,12 @@ void SynthNote::setFilterCutoff(float value)
         (value - 64.0f) * ctl.filtercutoff.depth / 4096.0f * 3.321928f;
 }
 
+void SynthNote::setBPM(int bpm_, int phase_)
```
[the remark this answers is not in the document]
JohannesLorenz: @friedolino78 You forgot that comment.

src/Synth/LFO.cpp
```diff
@@ -86,6 +73,28 @@ LFO::LFO(const LFOParams &lfopars, float basefreq, const AbsTime &t, WatchManage
 LFO::~LFO()
 {}
 
+void LFO::updatePars()
+{
+    waveShape = lfopars.PLFOtype;
+    int stretch = lfopars.Pstretch;
+    if(stretch == 0)
+        stretch = 1;
+    const float lfostretch = powf(basefreq / 440.0f, (stretch - 64.0f) / 63.0f);
```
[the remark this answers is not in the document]
JohannesLorenz: @friedolino78 You forgot that comment?

src/Synth/LFO.cpp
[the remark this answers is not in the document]
JohannesLorenz: You mean subtracting 0.25 before or after adding the start phase does not make a difference (commutative addition)?

src/Params/LFOParams.h
friedolino78: ok

src/Synth/LFO.cpp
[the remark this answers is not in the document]
friedolino78: exactly

src/Synth/SynthNote.cpp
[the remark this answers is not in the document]
friedolino78: it's a remainder from the first attempt including sync to midi HW. But that will be the second step, once this is merged.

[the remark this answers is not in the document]
friedolino78: I revert it

src/Synth/LFO.cpp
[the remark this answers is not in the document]
friedolino78: indeed. I even forgot, that I have removed this comment, at all. i put it back in.

src/Synth/LFO.cpp
```diff
@@ -173,44 +182,41 @@ void LFO::releasekey()
...
-    float out = baseOut(waveShape, phase);
-    // Apply intensity
+    // refresh freq if tempo has changed
+    if (lfopars.numerator && lfopars.denominator && tempo != time.tempo) {
+        tempo = time.tempo;
+        float lfofreq = float(tempo) * float(lfopars.denominator)/(240.0f * float(lfopars.numerator));
+        phaseInc = fabsf(lfofreq) * dt;
+    }
+
+    float out = baseOut(waveShape, fmod(phase + (lfopars.Pstartphase + 63.0f) / 127.0f, 1.0f));
```
JohannesLorenz, line 219: In the code below... doesn't this PR remove the `lfopars.Pstartphase` part from `phase`, which will make every use of `phase` below behave different? I'd expected that your would rather do `fmod(phase += ...`.

src/Synth/LFO.cpp
```diff
@@ -173,44 +182,41 @@ void LFO::releasekey()
...
-                lfointensity = powf(2, lfopars_.Pintensity / 127.0f * 11.0f) - 1.0f; // [0...2047] cent
+                lfointensity = powf(2, lfopars.Pintensity / 127.0f * 11.0f) - 1.0f; // [0...2047] cent
                 break;
         }
     }
     
-    float out = baseOut(waveShape, phase);
-    // Apply intensity
+    // refresh freq if tempo has changed
+    if (lfopars.numerator && lfopars.denominator && tempo != time.tempo) {
```
JohannesLorenz, line 213: Where is freq refreshed if only numerator or denominator change?

JohannesLorenz (commented):
Reviewed all code except EffectMgr.h/cpp (which I still don't understand).

src/Params/LFOParams.cpp
```diff
@@ -238,6 +247,8 @@ void LFOParams::defaults()
     Pcontinous  = Dcontinous;
     Pfreqrand   = 0;
     Pstretch    = 64;
+    numerator   = 0;
```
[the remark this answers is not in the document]
friedolino78: good idea

src/Params/LFOParams.h
```diff
@@ -69,6 +69,8 @@ class LFOParams:public Presets
         float         fadein; /**<fadein, relative to delay*/
         float         fadeout; /**<fadeout on key release (10.0=off)*/
         unsigned char Pcontinous; /**<1 if LFO is continous*/
+        int           numerator=0;  /**<numerator for integer ratio between system tempo and LFO freq (0=off)*/
```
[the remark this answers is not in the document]
friedolino78: right, seems like I was a bit overprotective when hunting the uninitialized variables

src/Synth/LFO.cpp
friedolino78: with this change adding Pstartphase is moved from once at init to cyclic. That's needed to tune the phase of two playing LFOs agains each other, or to tune the LFO phase to a synced delay.
+= works only if the startphase is only added once.

src/Synth/LFO.cpp
friedolino78: line 96 in updatePars()

JohannesLorenz: So numerator and denominator are part of `lfopars`, so they update `lfopars.last_update_timestamp`, and so are handled by `updatePars` line 196, but tempo is not part of `lfopars`, which means we have to additionally check it here?

src/Synth/LFO.cpp
JohannesLorenz: OK, agreed that the "+= works only if the startphase is only added once". However, `phase` is still a different value after this line than before the PR, so you change behavior in the rest of this function. Isn't this wrong? Instead, can't you do `const float phaseWithStartPhase = fmod ...` and then replace all further use of `phase` in this function by `phaseWithStartPhase`?

src/Synth/LFO.cpp
friedolino78: yes. tempo is in the global used time structure. if it's changed we refresh, but only if bpm sync is active (both numerator and denominator !=0)

src/Synth/LFO.cpp
friedolino78: you are right. at least for lines 221 and 292 the startphase has to be included.

friedolino78:
unfortunatelly my data drive crashed. I hope i can reconstruct the things i did yesterday without pushing them to repo.

src/Params/LFOParams.h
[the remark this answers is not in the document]
JohannesLorenz: I think this is still not solved?

src/Synth/LFO.cpp
```diff
@@ -289,7 +289,7 @@ float LFO::lfoout()
         computeNextFreqRnd();
     }
             
-    float watch_data[2] = {phase, out};
+    float watch_data[2] = {phaseWithStartphase, out};
```
JohannesLorenz, line 292: This will now take the phase without the increase from the previous lines. I think that's slightly wrong.

JohannesLorenz, line 292: Also, the above `if(phase >= 1)` will behave different now, because the start phase is not included? So this will not catch if `phaseWithStartphase >= 1`. It will also call `computeNextFreqRnd` later now.

src/Params/LFOParams.h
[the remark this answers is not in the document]
friedolino78: yeah, that fix went away with that defective hard drive.

src/Synth/LFO.cpp
friedolino78: it makes more sense, that watch shows the phase that was used to calculate out. So in a way it was wrong before.

friedolino78: if(phase >= 1) is only used to calculate a new random amplitude every cycle.
the exact phase where this happens doesn't matter so much because the value is lerped. The effort to put it to the time of the actual "phase reset" is quite high due to possible negative values of startphase.
long story short: i would keep it this way.

JohannesLorenz: OK, just to be sure, you have considered that `phase` could now be `>1` in `const float tmp = (incrnd * (1.0f - phase) + nextincrnd * phase);`? Is this acceptable?

friedolino78: phase cant be > 1 in that line, because it is limited in line 285 in the tick before

JohannesLorenz: Oh, right! resolving this comment.

JohannesLorenz (commented):
Reviewed all code except EffectMgr.h/cpp . No issues found .

src/Effects/EffectMgr.cpp
```diff
@@ -117,6 +120,149 @@ static const rtosc::Ports local_ports = {
...
+                    case 2:
+                        // invert:
+                        // delay = (Pdelay / 127.0f * 1.5f); //0 .. 1.5 sec
+                        Pdelay = (int)roundf((20320.0f / (float)eff->time->tempo) * 
+                                             ((float)eff->numerator / (float)eff->denominator));
+                        if (eff->numerator&&eff->denominator)
+                            eff->seteffectparrt(2, Pdelay);
+                        break;
+                    case 3:
+                    case 4:
```
fundamental, line 144: To make things more clear, it would be great to have a comment hinting what the various cases refer to (i.e. the subtype of effect)

src/Effects/EffectMgr.cpp
```diff
@@ -117,6 +120,149 @@ static const rtosc::Ports local_ports = {
...
+                if (val) {
+                    eff->denominator = val;
+                    int Pdelay, Pfreq;
+                    float freq;
+                    switch(eff->nefx) {
+                    case 2:
+                        // invert:
+                        // delay = (Pdelay / 127.0f * 1.5f); //0 .. 1.5 sec
+                        Pdelay = (int)roundf((20320.0f / (float)eff->time->tempo) * 
+                                             ((float)eff->numerator / (float)eff->denominator));
```
fundamental, line 211: It doesn't look like the divide by zero case is handled or validated in this code.

src/Effects/EffectMgr.cpp
```diff
@@ -117,6 +120,149 @@ static const rtosc::Ports local_ports = {
...
+                        // denominator = (tempo / 15.0) * (numerator/freq)
+                        value = ((float)eff->time->tempo / 15.0f) * ((float)eff->numerator / freq);
+                        break;
+                    case 1:
+                    case 6:
+                    case 7:
+                    default:
+                        break;
+                    }
+                d.reply(d.loc, "i", int(roundf(value)));
```
fundamental, line 263: The returned value does not appear to be the set value.
For OSC ports with 'parameter' semantics you should be able to run:

/port/name ,i 1234 and then /port/name will result in /port/name ,i 1234 getting returned. As such the reply value should be just eff->denominator, no?

src/Effects/EffectMgr.h
```diff
@@ -73,6 +73,10 @@ class EffectMgr:public Presets
         int     nefx;
         Effect *efx;
         const AbsTime *time;
+        
+        int numerator = 0;
```
fundamental, line 77: Should these values be stored in this class or in the child classes?

fundamental: Also if these values are to be stored here, are they being written to XML correctly? I'm not seeing code handling that in EffectMgr.cpp.

src/Effects/EffectMgr.cpp
friedolino78: good idea.

src/Effects/EffectMgr.cpp
friedolino78: hmm, maybe it's smarter to limit denominator to [1 ... 99]

src/Effects/EffectMgr.cpp
friedolino78: sounds plausible

src/Effects/EffectMgr.h
friedolino78: oh, that was missing. is being added now.

src/Effects/EffectMgr.cpp
friedolino78: now I also catch zero values,

src/Effects/EffectMgr.cpp
```diff
@@ -117,6 +120,98 @@ static const rtosc::Ports local_ports = {
...
+            }
+        }},
+    {"denominator::i", rShort("dem") rDefault(4) rLinear(1,99)
+        rProp(parameter) rDoc("Denominator of ratio to bpm"), NULL,
+        [](const char *msg, rtosc::RtData &d)
+        {
+            EffectMgr *eff = (EffectMgr*)d.obj;
+            if(rtosc_narguments(msg)) {
+                int val = rtosc_argument(msg, 0).i;
+                if (val) {
```
fundamental, line 176: I'd change this to if(val > 0) to prevent negative values (note the GUI will not send negative values, but the users can with the existing receiver code)

src/Effects/EffectMgr.cpp
```diff
@@ -117,6 +120,98 @@ static const rtosc::Ports local_ports = {
...
             }
         }},
+    {"numerator::i", rShort("num") rDefault(0) rLinear(0,99)
+        rProp(parameter) rDoc("Numerator of ratio to bpm"), NULL,
+        [](const char *msg, rtosc::RtData &d)
+        {
+            EffectMgr *eff = (EffectMgr*)d.obj;
+            if(rtosc_narguments(msg)) {
+                int val = rtosc_argument(msg, 0).i;
+                if (val) {
```
fundamental, line 130: The range is 0..99 right now, so val being zero would appear to be valid. As such it should be if(val >= 0). If the zero case isn't intentional feel free to modify the condition as needed.

fundamental:
Looks like the vast majority of comments have been addressed. There are now two minor comments left.

Of note, is something wrong with our unit tests? I don't see them running on this PR. I see a XML representation change without the corresponding change in the unit test, so that makes me suspect that this PR would change the tests from passing to failing.

src/Effects/EffectMgr.cpp
friedolino78: ok

src/Effects/EffectMgr.cpp
friedolino78: factor 0 makes no sense. maybe i change the range.

friedolino78: it makes only sense as default because the freq/delay is set via preset.

friedolino78:
> Looks like the vast majority of comments have been addressed. There are now two minor comments left.
> 
> Of note, is something wrong with our unit tests? I don't see them running on this PR. I see a XML representation change without the corresponding change in the unit test, so that makes me suspect that this PR would change the tests from passing to failing.

a manual quick test passed:
adding some fx with different bpm ratios
save instrument A
load another instrument B 
load saved instrument A 

but the EffectTest fails with segfault

fundamental:
I was able to identify why PluginTest wasn't failing. The main() function in that test never had a return value specified, so it was returning success regardless of what the test itself encountered. It looks like that should be the only failure now, though it's a trivial one to resolve. All other tests appear to have run fine on this branch based upon me running tests locally.

friedolino78:
> I was able to identify why PluginTest wasn't failing. The main() function in that test never had a return value specified, so it was returning success regardless of what the test itself encountered. It looks like that should be the only failure now, though it's a trivial one to resolve. All other tests appear to have run fine on this branch based upon me running tests locally.

i restored the return line. and the test results are:
```
48: Test timeout computed to be: 1500
48: ok 1 - 0.1f < sum...
48: not ok 2 - similar ints...
48: # Expected 157678, but observed 157822 instead (line 234)
48: not ok 3 - fdata == result...
48: # Failure on line 235
48: 4257 vs 4261
48: total cost = 8.000000
48: REF - <par name="route" value="0" />
48: NEW + <par name="route" value="0" />
48: NEW + </EFFECT>
48: REF - </EFFECT>
48: NEW + <par name="denominator" value="4" />
48: NEW + <par name="numerator" value="0" />
48: REF - <par name="route" value="0" />
48: NEW + <par name="route" value="0" />
48: NEW + </EFFECT>
48: REF - </EFFECT>
48: NEW + <par name="denominator" value="4" />
48: NEW + <par name="numerator" value="0" />
48: # 2 test(s) failed out of 3 (currently passing 33.333333% tests)
1/1 Test #48: PluginTest .......................***Failed   10.69 sec
```
i guess the missing route param was introduced in the commit where the plugin test was deactivated. Maybe it's best to clean it up there first.

friedolino78:
had to rebase. now functional plugintest passes.

friedolino78:
new rebase to master and squash into a single commit.

PR #119, opened 2021-03-30
Contribute Homebrew formula

kennystrawnmusic:
Found a way to build from source on a Mac via Homebrew; submitting the working formula for review (tested on M1 MacBook Pro, hence `make -j8` — if there's a way to use `sysctl -n hw.physicalcpu` instead without using `system` to invoke a shell, please update this.

### DIFF
--- a/homebrew.rb
+++ b/homebrew.rb
@@ -1,0 +1,43 @@
+# Documentation: https://docs.brew.sh/Formula-Cookbook
+#                https://rubydoc.brew.sh/Formula
+# PLEASE REMOVE ALL GENERATED COMMENTS BEFORE SUBMITTING YOUR PULL REQUEST!
+class Zynaddsubfx < Formula
+  desc "Fully featured musical software synthesizer for Linux, MacOS, BSD, and Windows"
+  homepage "https://zynaddsubfx.sf.net"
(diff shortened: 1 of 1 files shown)


JohannesLorenz:
Hi, thanks for the contribution! Should we just merge this as is? Or do you plan to add more commits?

Do you want to setup CI for this, so that the package is built automatically?

kennystrawnmusic:
> Hi, thanks for the contribution! Should we just merge this as is? Or do you plan to add more commits?
> 
> Do you want to setup CI for this, so that the package is built automatically?

Definitely merge as is, and optionally submit upstream to Homebrew/brew@master as well, since that will then allow other Mac users to `brew install --build-from-source zynaddsubfx` without any problems.

fundamental:
Looks like this is building a headless version and also it's using one of the zip files that github generates which continues to be terribly broken with respect to submodules (long standing issue that GitHub themselves is responsible for). As such I'm not in favor of merging this. Good to know we're building on M1 though.

So, if you have a more complete build working using the right sources I'm guessing that would go over in the zyn-fusion-build repo and that should be usable. Thanks for the interest and best of luck with a possible future PR.

PR #120, opened 2021-04-10
UI: Remove duplicate OscilGen prepare call

JohannesLorenz:
The phase/mag ports already call prepare, no reason to call it twice.

Checks done:

* Tested PAD and AD that change of the mag and phase sliders did affect all oscilloscopes + sound
* Added a printf into "prepare:b". It now appears half the time as on master.


fundamental:
Makes sense. That leaves the prepare calls in the more general refresh, redrawoscil, clear, and use-as-base. I'm guessing those are well handled as well, but I'm not 100% sure on that one. Removing the phase/mag ones should be totally safe though.

PR #121, opened 2021-04-10
Move prepare:b to where it has been called

JohannesLorenz:
Prior to this commit, OscilGen's prepare/magnitude/phase ports prepare
data and send it to prepare:b. Since we have no RT thread in OscilGen
anymore, this commit does not send anything to the RT side anymore -
instead, code where the new FFT data has been sent is replaced by the
content of the prepare:b port.

This fixes memory leaks, because prepare:b did not "/free" the freqs
prior to this commit.

### DIFF
--- a/src/Misc/MiddleWare.cpp
+++ b/src/Misc/MiddleWare.cpp
@@ -1062,19 +1062,19 @@ public:
                     const WaveTable* wt;
                     if(!params.wave_requests.size())
                     {
-                        Tensor1<WaveTable::float32>* nc_freqs; // non-constant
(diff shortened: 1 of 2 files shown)


fundamental:
Looks like the code behavior should be fine. Since the activity is occurring in the middleware thread you can eliminate the extra messages going back to the "/free" path. i.e. just delete [] within the callback.

JohannesLorenz:
Oh, already fixed the `delete[]`, just pushed to the wrong remote. Thanks for checking the code.

PR #122, opened 2021-04-25
add a nyquist factor to OscilGen::get

friedolino78:
implemented as ordered ;-)

### DIFF
--- a/src/Synth/OscilGen.cpp
+++ b/src/Synth/OscilGen.cpp
@@ -1058,11 +1058,14 @@ short int OscilGen::get(float *smps, float freqHz, int resonance, float nyquistF
 
     clearAll(outoscilFFTfreqs, synth.oscilsize);
 
-    int nyquist = (int)(0.5f * synth.samplerate_f / fabsf(freqHz) * nyquistFactor) + 2;
+    int nyquist = (int)(0.5f * synth.samplerate_f / fabsf(freqHz)) + 2;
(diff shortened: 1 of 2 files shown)


src/Synth/OscilGen.cpp
```diff
@@ -1059,18 +1059,21 @@ short int OscilGen::get(float *smps, float freqHz, int resonance)
...
+        nyquistLimited = synth.oscilsize / 2;
 
     //Process harmonics
     {
         int realnyquist = nyquist;
 
         if(Padaptiveharmonics != 0)
             nyquist = synth.oscilsize / 2;
-        for(int i = 1; i < nyquist - 1; ++i)
+        for(int i = 1; i < nyquistLimited - 1; ++i)
```
JohannesLorenz, line 1076: `Padaptiveharmonics` has no effect anymore now.

friedolino78: > `Padaptiveharmonics` has no effect anymore now.

it is also used inside OscilGen::adaptiveharmonic()
look at line 913

JohannesLorenz: What I really meant was that the 2 lines above this line do not have an effect on this for loop anymore. Which means that this code breaks the current behavior.

JohannesLorenz:
I remember that I asked for this because we need AA for the wavetable branch, but is this really the right tool for that job? (Didn't you even write a PR which does the AA?)

friedolino78:
> I remember that I asked for this because we need AA for the wavetable branch, but is this really the right tool for that job? (Didn't you even write a PR which does the AA?)

yes, this one can be dumped.

PR #123, opened 2021-04-25
Bump C++ version to C++14

JohannesLorenz:
For use of `CMAKE_CXX_STANDARD`, this commit raises CMake's version from
3.0 to 3.1.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -48,7 +48,7 @@ else()
 endif()
 
 CHECK_FUNCTION_EXISTS(sched_setscheduler HAVE_SCHEDULER)
-set(CMAKE_REQUIRED_FLAGS "-std=c++11")
+set(CMAKE_REQUIRED_FLAGS "-std=c++14")
 CHECK_CXX_SOURCE_COMPILES(
     "#include <future>
     #if ATOMIC_INT_LOCK_FREE<2
@@ -262,14 +262,15 @@ set (BuildOptions_SSE
 
 check_cxx_compiler_flag("${BuildOptions_SSE} -Werror" SUPPORT_SSE)
 
+set (CMAKE_CXX_STANDARD 14)
 set (BuildOptionsBasic
-    "-std=c++11 -Wno-unused-parameter -O3 -ffast-math -fomit-frame-pointer"
+    "-Wno-unused-parameter -O3 -ffast-math -fomit-frame-pointer"
     CACHE STRING "basic X86 compiler options"
 )
 STRING(APPEND BuildOptionsBasic " ${BuildOptions_ExtendedWarnings}")
 
 set (BuildOptionsDebug
-    "-std=c++11 -O0 -g3 -ggdb -Wall -Wno-unused-parameter -Wpointer-arith"
+    "-O0 -g3 -ggdb -Wall -Wno-unused-parameter -Wpointer-arith"
     CACHE STRING "Debug build flags")
 STRING(APPEND BuildOptionsDebug " ${BuildOptions_ExtendedWarnings}")
 
--- a/src/Tests/CMakeLists.txt
+++ b/src/Tests/CMakeLists.txt
@@ -6,6 +6,7 @@ endfunction()
 #for tests looking for files stored in the source dir
 add_definitions(-DSOURCE_DIR="${CMAKE_CURRENT_SOURCE_DIR}")
 
+CXXTEST_ADD_TEST(CppTest CppTest.cpp ${CMAKE_CURRENT_SOURCE_DIR}/CppTest.h)
 CXXTEST_ADD_TEST(ControllerTest ControllerTest.cpp ${CMAKE_CURRENT_SOURCE_DIR}/ControllerTest.h)
 CXXTEST_ADD_TEST(EchoTest EchoTest.cpp ${CMAKE_CURRENT_SOURCE_DIR}/EchoTest.h)
 #CXXTEST_ADD_TEST(SampleTest SampleTest.h)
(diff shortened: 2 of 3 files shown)


JohannesLorenz:
CI passed, except of the usual build failures. PR is ready from my side.

fundamental:
I'd like to avoid ongoing CI failures, so I'm trying to see if I can resolve it now. If I had more time floating around I'd be tempted to shift from cxxtest to an internal testing approach (similar to rtosc), given the number of cxxtest install issues that have occurred over the years and our limited testing complexity.

fundamental:
Ugh. Upstream has bad python3 handling code, which results in bad python3 adapters getting called which results in the generator never working. Upstream is totally dead as well. Working on a fix....

falkTX:
I would like to ask for the C++ spec version to remain as C++11 please.
C++11 is the last one that can be used with good old 4.9.x gcc.

The new standards are not exactly great for audio, since they tend to do a lot of hidden stuff behind the scenes which is hard to ensure that is is RT safe.

fundamental:
@falkTX Since you're much more qualified to talk about packaging concerns, any specific CMake version we should limit ourselves to? Right now the minimum acceptable version is ancient, so I wouldn't mind pushing that up myself.

falkTX:
cmake is fine.
it is much easier to backport cmake than it is a whole compiler toolchain.

falkTX:
to be clear, I mean at least from my side I have no specific requiremends on cmake.

you can see the available cmake versions in ubuntu at https://launchpad.net/ubuntu/+source/cmake
I would say based on that, that a limit on maximum 3.10 would be appreciated.

fundamental:
Based upon the comments I'll reject the upgrade for now. Feel free to upgrade CMake versions though. With newer cmake it should be easy to specify C++ standard which should work out well if you wanted C++14 for internal development purposes. Semi-related, looks like we should no longer depend on cxxtest, so that's one less thing to fuss with. (I'm hoping that the CI server won't disagree with me on that point in about 3 minutes time).

JohannesLorenz:
> The new standards are not exactly great for audio, since they tend to do a lot of hidden stuff behind the scenes which is hard to ensure that is is RT safe.

@falkTX Do you say that versions beginning from C++14 can not be used in realtime context at all? What "hidden stuff" are you refering to?

falkTX:
not saying that, just that it is easier to have unexpected things due to how much stuff happens behind the scenes.
starts with the "auto" var type, which can give you a pointer, or a shared reference or regular type.

lambas is something that comes to mind that it is not clear exactly what happens.
can they be used on a realtime context? (might be useful for similar math calculations)

I am being a bit too pessimistic here, but we have the same exact situation with for example Qt.
Can we safely use Qt classes inside RT code? It certainly has nice things for mutexes, semaphores and others useful for audio.
But we can't just assume they are RT safe, as they are not created with that in mind.
Plus, an update to Qt might change behaviour that was fine for RT before.

I see the new C++ in a similar fashion.
It can be annoying and more verbose to do things "the old way" or alike C code, but IMO those are more reliable in the end.

JohannesLorenz:
@falkTX I think you are confusing things. No offense intended in this post, but I want to clarify this.

Zyn is already C++11 for >5 years, and we already use Lambdas/auto almost everywhere in the RT code (without problems). I googled a bit: [A lambda isn't anything special in C++11. It's an object like any other object. A lambda expression results in a temporary, which can be used to initialize a variable on the stack](https://stackoverflow.com/a/12203426/985292). From `auto`, I recall that it gives you the exact type that you feed it (i.e. no casts), unless specified explicitly.

Moreover: The fact that new features like lambda/auto/... can do "a lot of hidden stuff behind the scenes" does not mean that upgrading your compiler forces your devs to use these features. You can use gcc-10.2 if you like, but still restrict your project to be C++03 compliant (`-std=...`).

I can be wrong and may oversee dangers, but I really can not think of any. Unless you give us a good example, I think this boils down to that you have a project which uses zyn, and your project does not work with anything >gcc-4.9? If this is the only reason, can't you just port your project to newer compilers? I mean, we can't stay at C++11 for the next 100 years... (again, no offense, but how long should we wait?)

Btw: you can check your program to be RT safe using [stoat](https://github.com/fundamental/stoat). It will bash you with errors about Qt, but at least for Zyn, you won't get errors for lambdas or similar features.

falkTX:
I am mostly worried about new things that we dont understand very well yet and might be subject to change.
To be more precise, "fear" that how these things are handled could be compiler-dependent.
C++11 can dictate that lambdas use the local stack, but future C++ versions might go with heap allocation instead.

My thoughts on the over-use of auto.. there can be functions can return std pointers (like shared_ptr or unique_ptr) which you simply do not directly see. Like for example a function that returns a heap allocated variable encapsulated on a shared_ptr. This will allocate and deallocate on the current function scope.

I have some things that use old compilers yes, but not strictly my own projects, more of a setup for building as generic of a binary as we can get. This is not by itself a good reason to stick with C++11, I agree with you there.

JohannesLorenz:
An old topic, but I could need C++17 now, especially that in order to solve #339 I want to use `std::filesystem` instead of adding 100s of Windows specific file access code (and there have been enough other cases).

So, here are some new arguments:

1. C++17 is already 8 years old]
2. Vital and Helm use C++14
3. Ardour uses C++17
4. [I posted some concerns on SO](https://stackoverflow.com/questions/79793849/how-do-real-time-audio-projects-deal-with-new-c-standards), and the comments directly at the OP speak for themselves

I propose opening a C++17 PR and move on here, but before, I would like to ask @fundamental to confirm that project decision.

PR #124, opened 2021-05-18
add sallen-key lp filter to oscilgen

friedolino78:
add a more analog style lp filter based on the sallen-key transfer function to oscilgen lpsk2
params are cutoff and resonance.

### DIFF
--- a/src/Synth/OscilGen.cpp
+++ b/src/Synth/OscilGen.cpp
@@ -22,6 +22,7 @@
 #include <cmath>
 #include <cstdio>
 #include <cstddef>
+#include <complex>
 
 #include <unistd.h>
 
@@ -29,6 +30,7 @@
 #include <rtosc/port-sugar.h>
 
 namespace zyn {
+
 
 #define rObject OscilGen
 const rtosc::Ports OscilGen::non_realtime_ports = {
@@ -70,7 +72,7 @@ const rtosc::Ports OscilGen::non_realtime_ports = {
             "Shape of distortion to be applied"),
     rOption(Pfiltertype, rShort("filter"), rOptions(No Filter,
             lp, hp1, hp1b, bp1, bs1, lp2, hp2, bp2, bs2,
-            cos, sin, low_shelf, s), rDefaultId(No Filter), "Harmonic Filter"),
+            cos, sin, low_shelf, s, lpsk2), rDefaultId(No Filter), "Harmonic Filter"),
     rParamZyn(Pfilterpar1, rShort("p1"), rDefault(64), "Filter parameter"),
     rParamZyn(Pfilterpar2, rShort("p2"), rDefault(64), "Filter parameter"),
     rToggle(Pfilterbeforews, rShort("pre/post"), rDefault(false),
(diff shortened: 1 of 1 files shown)


friedolino78:
formular from: [https://electronics.stackexchange.com/questions/243122/low-pass-filter-and-gain-calculation](https://electronics.stackexchange.com/questions/243122/low-pass-filter-and-gain-calculation)

fundamental:
I like the function, but I'm much less sure about the parameter mapping. For the cutoff frequency, shouldn't low be low cutoff and high high cutoff? I think that's how other controls work. Per Q, it does seem like the low end is well below useful Q values. Perhaps plot out the function with something like python+matplotlib to see what I'm looking at.

friedolino78:
> I like the function, but I'm much less sure about the parameter mapping. For the cutoff frequency, shouldn't low be low cutoff and high high cutoff? I think that's how other controls work. Per Q, it does seem like the low end is well below useful Q values. Perhaps plot out the function with something like python+matplotlib to see what I'm looking at.

in my system cutoff is proportional to P1. but it didn't like zero. So I raised the lower param end.
same for par2 (Q)
to make it more agressive i made it 3rd order, too.

fundamental:
2021-06-lp4
2021-06-lp3
2021-06-lp2
2021-06-lp1

Seems like par2 might overemphasize loads of high Q resonance, but that's part of the character of the desired filter. I'd still think that par1=0 would be a low cutoff and par1=1.0 would be basically passing the input unaltered. Is that inconsistent with the other filters? I only plotted your code with some quickly put together matplotlib style plots.

friedolino78:
> Seems like par2 might overemphasize loads of high Q resonance, but that's part of the character of the desired filter. I'd still think that par1=0 would be a low cutoff and par1=1.0 would be basically passing the input unaltered. Is that inconsistent with the other filters? I only plotted your code with some quickly put together matplotlib style plots.

looking to the code you are right. but in the gui par1=1.0 passes all. somewhere is another y=1-x in the parameter path. i don't know why.

fundamental:
I've been running the code directly instead of checking with the GUI, if there's another flip such that for the parameter knob, full clockwise is all pass and full counterclockwise is almost full cutoff, then we're all set.

friedolino78:
> I've been running the code directly instead of checking with the GUI, if there's another flip such that for the parameter knob, full clockwise is all pass and full counterclockwise is almost full cutoff, then we're all set.

line 567:    const float par    = 1.0f - Pfilterpar1 / 128.0f;

fundamental:
Final check, given the above figures, any desire to change the par2 mapping onto Q? I'm fine merging or letting you tweak it if you desire.

friedolino78:
> Final check, given the above figures, any desire to change the par2 mapping onto Q? I'm fine merging or letting you tweak it if you desire.

now that you made the nice figures i will look for it tomorrow.

friedolino78:
I changed the parametrization a bit. I guess that's it now.

src/Synth/OscilGen.cpp
```diff
@@ -1864,12 +1864,12 @@ FILTER(s)
...
 }
 
-FILTER(lpsk2)
+FILTER(lpsk)
 {
     float tmp2PIf = 2.0f * PI * (1.05f-par)*64.0f;
     std::complex<float> s  (0.0f,2.0f*PI*i);
     float vOut = tmp2PIf * tmp2PIf;
-    std::complex<float> vIn = s*s + tmp2PIf*s/(4*par2+0.5f) + tmp2PIf*tmp2PIf;
+    std::complex<float> vIn = s*s + tmp2PIf*s/((par2)+(2.0f*par*par2)+0.5f) + tmp2PIf*tmp2PIf;
```
fundamental, line 1872: Just a sanity check you do want par*par2 and not par2*par2, right? I would have just merged, but right now the PR doesn't compile since you changed lpsk2 to lpsk without updating some other glue code (just find the osc_lpsk2 line or check the CI failure)

friedolino78: par*par2 is right because for higher freqs (par) the impact of q (par2)  should be higher(3x).
osc_lpsk2 fix is underway.

fundamental:
Merged :)

PR #126, opened 2021-05-27
add comb filter

friedolino78:


src/DSP/CombFilter.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+void CombFilter::setfreq_and_q(float frequency, float q_)
+{
+    setfreq(frequency/sr);
+    setq(q_);
+}
+
+void CombFilter::setfreq(float ff)
+{
+    delayfwd = sr/ff;
```
fundamental, line 71: Boundaries need to be added to ensure no out of bounds accesses can occur.

friedolino78: ok

friedolino78:
for this one a float LFO depth value could be good

src/DSP/CombFilter.cpp
```diff
@@ -16,7 +16,7 @@ CombFilter::CombFilter(Allocator *alloc, unsigned char Ftype, float Ffreq, float
     unsigned int srate, int bufsize)
     :Filter(srate, bufsize), memory(*alloc), sr(srate), gain(1.0f), type(Ftype)
 {
-    mem_size = sr/30;
+    mem_size = sr/24;
```
fundamental, line 19: Seems like this should be buffer_size+maximum_delay_lookback if I'm reading the rest of the code correctly. I think with the current value it is assuming something like buffer_size caps out at around 128 with 48kHz sample rate. (Not a terrible place to put it, but give users the choice and they'll break the assumption).

src/DSP/CombFilter.cpp
```diff
@@ -95,8 +86,9 @@ void CombFilter::setfreq_and_q(float frequency, float q)
     setq(q);
 }
 
-void CombFilter::setfreq(float ff)
+void CombFilter::setfreq(float freq)
 {
+    float ff = limit(freq, 25.0f, 40000.0f);
```
fundamental, line 91: Excellent :)

src/DSP/CombFilter.cpp
friedolino78: > Seems like this should be buffer_size+maximum_delay_lookback if I'm reading the rest of the code correctly. I think with the current value it is assuming something like buffer_size caps out at around 128 with 48kHz sample rate. (Not a terrible place to put it, but give users the choice and they'll break the assumption).

the buffer gets accessed via
(mem_size-buffersize+i)-delayfwdbuf[i])
this index has to be > 0 for linear sampling and >1 for 3rd order sampling
while
delayfwd = sr/ff
(mem_size-buffersize+i)-sr/ff > 1
(mem_size-buffersize+i) > sr/ff + 1
mem_size >  sr/ff + 1 + buffersize -i
minimal ff: 25
so i would say:
mem_size = sr/24 + buffersize
is safe under all circumstances

fundamental:
```
(mem_size-buffersize+i)-delayfwdbuf[i] must be >= 0
minimum value with i=0
delayfwdbuf is maximal at samplerate/25.0
so mem_size-buffersize-samplerate/25.0 >=0
mem_size >= samplerate/25.0+buffersize
```

sr/24+buffersize also works, though I'm likely missing something with the 25 vs 24 difference.

friedolino78: > mem_size >= samplerate/25.0+buffersize
> ```
> 
> sr/24+buffersize also works, though I'm likely missing something with the 25 vs 24 difference.

it was just a safety margin due to truncation of the division result sr/24 and a few samples more for higher order interpolation. but maybe adding a few samples is more straight forward.

fundamental: Fair enough for me.

src/DSP/Filter.cpp
```diff
@@ -52,6 +53,10 @@ Filter *Filter::generate(Allocator &memory, const FilterParams *pars,
             if(filter->outgain > 1.0f)
                 filter->outgain = sqrt(filter->outgain);
             break;
+            filter->setgain(pars->getgain());
+            break;
+        case 3:
+            filter = memory.alloc<CombFilter>(&memory, Ftype, 1000.0f, pars->getq(), srate, bufsize);
```
fundamental, line 59: Missing break

fundamental:
Running the current code crashes and without modification does not even spawn a comb filter. WIP diff is attached, though there is some additional memory fault occurring within the filter.
[df.txt](https://github.com/zynaddsubfx/zynaddsubfx/files/6594927/df.txt)

fundamental:
Took another look at the code. Looks like the memory issues are due to an implicit assumption that sizeof(float) == 1. So there's a few spots where you need to add a ```*sizeof(float)``` which I think will solve your crashes.

friedolino78:
everything seems to work now but the frequency mapping is wrong. a noise burst with 100% Q and 100% f.track creates nice tones but playing only octaves for example from lowest to highest c, sounds like a logarithmic mapping. as if it's proportional midi note number.

fundamental:
Oh nice, sounds like some major progress from last time I checked the branch.

friedolino78:
> Oh nice, sounds like some major progress from last time I checked the branch.

thx. if only that freq tuning problem wasn't there. it is against all i know of physics and signal processing.
if you have any idea, let me know.

fundamental:
Yeah, the frequency tracking certainly feels wrong to me running the code.

- Comb filter
- Q=max
- freq=max
- freq-tracking=max
- oscillator=pink noise

Using baudline's primary Hz we should be able to measure the first harmonic pretty consistently if the noise source has more energy on the lower freqs (i.e. pink noise).
Going up in octaves: (normally I'd expect doubling of freq
- 667 Hz
- 911 Hz
- 1130 Hz
- 1285 Hz
- 1385 Hz
- 1395 Hz

fundamental:
Not sure about the full solution, but I found the bug which causes loud noises for the feedback path. You're grabbing values from output, but you're not updating the output buffer for each individually output sample. So if you end up with delay<buffersize then you do not grab an accurate past sample.

Reading https://en.wikipedia.org/wiki/Comb_filter the math looks right for the most part, though I feel like some sort of factor of two might be missing. The behavior of the feedback mode is certainly going to change when you fix the aforementioned bug.

friedolino78:
> Not sure about the full solution, but I found the bug which causes loud noises for the feedback path. You're grabbing values from output, but you're not updating the output buffer for each individually output sample. So if you end up with delay<buffersize then you do not grab an accurate past sample.
> 
> Reading https://en.wikipedia.org/wiki/Comb_filter the math looks right for the most part, though I feel like some sort of factor of two might be missing. The behavior of the feedback mode is certainly going to change when you fix the aforementioned bug.

great work. that updating of the output buffer solved the freq mapping.

friedolino78:
profiling shows the combfilter performance lower than moog but slightly higher than analog.
most consuming is the resampling.
[https://imgur.com/a/fqXFF5r](https://imgur.com/a/fqXFF5r)

fundamental:
With that bug fixed then this PR should be good to go. I know previously you could get 1e100 outputs of the system (disabling the tanh), but that was a result of the broken backwards path. If there are no further proposed changes I'd say merge away :+1:

PR #127, opened 2021-06-01
Fixed the documentation for a few OSC endpoints.

pgervais:
The fix was to add a missing call to rDoc. Also documented the /quit endpoint.


fundamental:
Looks good :+1:

PR #134, opened 2021-08-26
add step sequencer to addsynth

friedolino78:
based on the internal BPM sync feature this branch adds a sequencer SEQ as modulator parallel to ENV and LFO


friedolino78:
example using SEQ to modulate comb filter cutoff frequency. 
[https://soundcloud.com/friedolino78/filterseq](https://soundcloud.com/friedolino78/filterseq)

PR #136, opened 2021-08-29
fix of the AnalogFilter Punch

friedolino78:
attempt to fix https://github.com/zynaddsubfx/zyn-fusion-issues/issues/289


fundamental:
I suspect the identification of the bug is right, but this behavior is still possibly wrong? If it's a problem of the filter smoothing starting at wrong value X and transitioning to Y, then it should start at Y if the user provided it at some point. I'm having an off day, so I might be reading this commit wrong, but it reads like it's doing the same behavior, just in a different routine.

fundamental:
I think changing the freq_smoothing.reset(freq*MAX_FREQ_CO) to a freq_smoothing.reset(freq) in the old location sorts the issue, but I'm just reading the code not executing it.

fundamental:
Hm, just read the comment on the other issue. I think you're right in indicating the starting frequency field isn't populated correctly. Perhaps updating on the first setfreq()?

friedolino78:
> I think changing the freq_smoothing.reset(freq*MAX_FREQ_CO) to a freq_smoothing.reset(freq) in the old location sorts the issue, but I'm just reading the code not executing it.
 
i guess you are wrong. the actual calculation of smoothing also uses the factor:
freq_smoothing.apply( freqbuf, buffersize, freq * MAX_FREQ_CO )

but the bad readability of the code could be a sign that a bit of refactoring is recommended.
the code redundancy of singlefilterout_freqbuf and singlefilterout and the cascaded multiplications with MAX_FREQ_CO and MAX_FREQ show improvement potential.

friedolino78:
> Hm, just read the comment on the other issue. I think you're right in indicating the starting frequency field isn't populated correctly. Perhaps updating on the first setfreq()?

setfreq would be indeed a better place as in the current draft because it gets less calls.

fundamental:
I can confirm the first commit does appear to resolve the issue. I wasn't expecting to see quite the number of changes in the later commits so, I guess I'll have to spend some more time reviewing them to make sure it doesn't change any other behavior.

I will say that at first pass it looks like the introduction of another fabsf(current_state-target) < threshold appears to be wrong. Value_Smoothing_Filter::apply() already had a chunk of code similar to that with a hardcoded threshold. I'll let you know when a full review is complete.

friedolino78:
> I can confirm the first commit does appear to resolve the issue. I wasn't expecting to see quite the number of changes in the later commits so, I guess I'll have to spend some more time reviewing them to make sure it doesn't change any other behavior.
> 
> I will say that at first pass it looks like the introduction of another fabsf(current_state-target) < threshold appears to be wrong. Value_Smoothing_Filter::apply() already had a chunk of code similar to that with a hardcoded threshold. I'll let you know when a full review is complete.

Sorry. But I couldn't endure the redundant code in the duplicated function and the "smoothing with 8x oversampling". 
I think the fixed threshold was the reason for the forth and back multiplication of MAX_FREQ.
A variable thresh seems to be the more elegant solution. Now I changed it a bit to reuse the existing threshold check you mentioned.

fundamental:
After some double checking, running on my local machine, plugging values into new/old code, etc, everything looks good to me. The unit test values need to be updated and I solved the mystery of why it looked like they weren't updated last time. If you use git log --follow FILE_NAME on the unit tests you'll see ca347aa1112d5e521d1f3bed62b871a28a3d26aa . The tests were renamed when the cxxtest dependency was removed and I totally spaced out, forgetting that git log wouldn't by default follow through renaming.

So, let's update those unit test values, squash, merge, and celebrate that another bug is dead :)

PR #137, opened 2021-08-30
make Compressor in Audio Output optional

friedolino78:
in this branch the StereoCompressor in the audio output will be changed to be optional an per default off to keep compatibility.

### DIFF
--- a/src/Nio/OutMgr.cpp
+++ b/src/Nio/OutMgr.cpp
@@ -127,7 +127,6 @@ string OutMgr::getSink() const
 
 void OutMgr::setAudioCompressor(bool isEnabled)
 {
-    printf("OutMgr - setting isOutputCompressionEnabled to: %d\n", isEnabled);
     currentOut->isOutputCompressionEnabled=isEnabled;
 }
 
--- a/src/Nio/OutMgr.cpp
+++ b/src/Nio/OutMgr.cpp
@@ -125,6 +125,17 @@ string OutMgr::getSink() const
     return "ERROR";
 }
 
+void OutMgr::setAudioCompressor(bool isEnabled)
+{
+    printf("OutMgr - setting isOutputCompressionEnabled to: %d\n", isEnabled);
+    currentOut->isOutputCompressionEnabled=isEnabled;
+}
+
+bool OutMgr::getAudioCompressor(void)
+{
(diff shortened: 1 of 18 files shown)


friedolino78:
local test results are fine. maybe a problem with a different compiler?

src/Nio/OutMgr.cpp
```diff
@@ -125,6 +125,17 @@ string OutMgr::getSink() const
     return "ERROR";
 }
 
+void OutMgr::setAudioCompressor(bool isEnabled)
+{
+    printf("OutMgr - setting isOutputCompressionEnabled to: %d\n", isEnabled);
```
fundamental, line 130: Good for development, but this should be removed for the final merging.

src/Nio/OutMgr.h
```diff
@@ -49,8 +49,11 @@ class OutMgr
         std::string getDriver() const;
 
         bool setSink(std::string name);
-
+
```
fundamental, line 52: Nit pick, but extra whitespace.

fundamental:
Looks like the compiler issue may likely be related to the stub implementation of Nio for when plugins are built. See the end of the src/Plugin/ZynAddSubFX/ZynAddSubFX.cpp for the one stub impl I could recall.

src/Nio/OutMgr.cpp
friedolino78: ok

friedolino78:
> Looks like the compiler issue may likely be related to the stub implementation of Nio for when plugins are built. See the end of the src/Plugin/ZynAddSubFX/ZynAddSubFX.cpp for the one stub impl I could recall.

it helped. i did it in also in DSSI. somehow they differ a little bit (waveEnd) maybe you have another look whether everything is correct.

fundamental:
It should be good if everything is building. One modification that I thought about today is saving the user's choice in the global settings xml rather than forcing them to enter in their choice every start. Opinions?

friedolino78:
> It should be good if everything is building. One modification that I thought about today is saving the user's choice in the global settings xml rather than forcing them to enter in their choice every start. Opinions?

it's a good idea.

fundamental:
Thanks for the fix :+1:

PR #139, opened 2021-09-05
fix latchmode initialization

friedolino78:
should fix https://github.com/zynaddsubfx/zyn-fusion-issues/issues/297


fundamental:
Looks good to me. I should have caught it last round, but oversights happen.

PR #140, opened 2021-09-14
ENVvStretch

friedolino78:


fundamental:
The changes all make sense reading through them. Did I miss a corresponding GUI PR such that I can try this out to get a feel for it?

PR #142, opened 2021-09-22
Voice limit

polluxsynth:
Here's my attempt at adding a settable voice limit per part (Pvoicelimit), similar to Pkeylimit, except it really limits the number of voices sounding not just the number of notes actually being played. This way, sounds with long release times won't cause xruns when playing lots of notes. In effect, the part will behave like a classic synthesizer with a fixed number of voices.

Caveat: voices which are silenced are entombed rather than killed, meaning that for a brief period after stealing voices there will be more voices playing than the voice limit governs. I felt this was still preferable to killing the voices as that results in audible clicks, and the voice limit is not a hard DSP limit anyway.

Effort has been made to select the most appropriate note to steal every time

### DIFF
--- a/src/Containers/NotePool.cpp
+++ b/src/Containers/NotePool.cpp
@@ -29,6 +29,18 @@ enum NoteStatus {
     KEY_LATCHED                = 0x04
 };
 
+const char *getStatus(int status)
+{
+    switch((enum NoteStatus)(status&NOTE_MASK))
+    {
+        case KEY_OFF:                     return "OFF ";
+        case KEY_PLAYING:                 return "PLAY";
+        case KEY_RELEASED_AND_SUSTAINED:  return "SUST";
+        case KEY_RELEASED:                return "RELA";
+        case KEY_LATCHED:                 return "LTCH";
+        default:                          return "INVD";
+    }
+}
 
 NotePool::NotePool(void)
     :needs_cleaning(0)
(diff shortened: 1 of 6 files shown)


fundamental:
Looks like I needed to click on some button to get CI to run through. Right now two tests are failing. Plugin Test which fails everytime the xml serialization changes (to avoid unintentional changes). It prints out a handy diff for you if you run it and in the case of this PR it should just be adding in the newly introduced key limiting value. The KitTest also fails which isn't too surprising since it's in the keylimit test which likely is getting different results from changing releasing into entombing.

Per the note stealing priorities, I like it. I was running it locally and it seems to identify age just fine, everything is wired into the GUI properly, The other types of age seem to be handled correctly as well. Targeting voice stealing the current note id seems like a reasonable approach as well.

I'd say fix up the tests and also change any introduced tab characters into spaces, then it should be mergable. Thanks for the work in tidying up this code.

polluxsynth:
Thanks for your comments. I'll have a look at this and fix up the tests and remove the tab characters. (I've been doing the development work on a newly installed machine and haven't done much customization).

polluxsynth:
Updated branch:
- Replace tabs with spaces.
- Fix failing tests.
- Add test in KitTests for voice limit function.

fundamental:
Excellent :) I love having features with new tests. Since this is just a cleanup of the previous patch set and tests are passing I just gave things a read through. Now to merge the zest side. :+1:

polluxsynth:
Thanks for merging this!

PR #143, opened 2021-09-25
Avoid PAD samples leaking memory (#291)

polluxsynth:
Return samples to non-realtime thread for deallocation when new
samples are installed.

### DIFF
--- a/src/Misc/MiddleWare.cpp
+++ b/src/Misc/MiddleWare.cpp
@@ -197,6 +197,8 @@ void deallocate(const char *str, void *v)
         delete (SclInfo*)v;
     else if(!strcmp(str, "Microtonal"))
         delete (Microtonal*)v;
+    else if(!strcmp(str, "PADsample"))
+        delete[] (float*)v;
     else
         fprintf(stderr, "Unknown type '%s', leaking pointer %p!!\n", str, v);
 }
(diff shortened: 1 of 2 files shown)


fundamental:
Close, but not quite. If it's a float vector then it should be delete[], not delete, no? Also looks like there's a stray tab character in there.

polluxsynth:
Damn, I really need to fix my editor configuration. The tab characters often show up in the diffs when I'm reviewing and I replace with spaces them manually but that's not the proper way to handle them obviously.

As for delete[], yes, it makes sense, I actually tried both, and it made no practical difference, I think this is because what delete[] does is invoke the destructor for each element in the vector, but with simple data types such as float, there is none (or it's a dummy) because there's nothing that needs to be done on the individual elements in order to free them. But of course it doesn't hurt, and makes more sense semantically since it is in fact a vector that is being freed.

My rationale for using a simple delete was that I figured that when the OSC message is received, the blob is just an anonymously allocated block of memory, and hence it should just be freed with a simple delete. But I've honestly not had a look at how blobs are passed; I was thinking that when using OSC over a socket, a blob would not have a specific type, but perhaps that's not the case here when the message passing is done within the application.

fundamental:
Did you intend to close this PR as well as one other PR?

polluxsynth:
Yeah, this got a bit messy. Yes, this one should be closed.

PR #145, opened 2021-09-26
Avoid PAD samples leaking memory (#291)

polluxsynth:
Return samples to non-realtime thread for deallocation when new
samples are installed to avoid old unused samples leaking memory.

Fix suggested by Mark McCurry.


polluxsynth:
Changed since last PR:
Converted tabs to spaces, slightly adjusted commit message.

fundamental:
Looks good to me. Thanks for the bugfix for a long standing omission. :+1:

polluxsynth:
Without your guidance I wouldn't have had a clue where to start looking at this...

PR #147, opened 2021-10-06
Fix delaying portamento bug (#304)

polluxsynth:
Resetting portamento.x to 0 at the start of initportamento() causes
the actual portamento process to restart from zero every time a new note
is played - for the note that is currently exhibiting portamento
(there is only one per voice (or actually, per Controller)). The
result is that subsequently played note(s) will not have portamento
until the ongoing restarted portamento cycle has completed.

Fix this by simply not resetting portamento.x until such time that
initportamento() has decided to actually initate a portamento cycle.

### DIFF
--- a/src/Params/Controller.cpp
+++ b/src/Params/Controller.cpp
@@ -275,8 +275,6 @@ int Controller::initportamento(float oldfreq_log2,
                                float newfreq_log2,
                                bool legatoflag)
 {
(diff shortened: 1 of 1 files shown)


polluxsynth:
Sure. I'm continuing to work on having individual portamento per note, so this is only a first step which only fixes the immediate bug.

PR #150, opened 2021-10-23
Per note portamento+fixes

polluxsynth:
Here, finally, is a PR for the per-note-portamento feature/fix.

After some thought and experimentation, I opted to squash the majority of the patches into a single commit, as I feel keeping the individual commits makes it rather confusing as there are things like the portamento class which were moved around between several locations as development went on, and I think it's fairly easy to follow what's been done even though the intermediate commits are not individually available.

The main commit is preceded by two, I would say, bugfixes which are not really part of the main feature, including https://github.com/zynaddsubfx/zynaddsubfx/pull/147 . The final commit adds tests for various portamento settings, which I felt made sense to keep separate, as some of them test things such as the threshold operation, which were not tested before.

Since portamento is now handled by a separate class, I created a separate PortamentoTest for it, which essentially is a rewrite of ControllerTest, which indeed only tested this feature. So the ControllerTest is now empty, but I left it in the test suite nevertheless, as a placeholder for future controller tests.


fundamental:
Thanks for bearing with the delay. Playing around with the branch everything is working more or less as described in the previous issue, so doing excellent on that front :)

Per modifications:

1. I'm not sure if this is intended or not. (I suspect it is), but portamento in legato+mono modes will initiate a portamento from a released note rather than just the currently actively held note. Different notes interact with envelope frequency based stretching (I think) which can result in some weirdness where a consistent arpegiation (releases the last note before playing the next) can result in partial portamento on the down swing. Fairly minor, but I would have expected the released notes to not trigger a portamento. Your call entirely though as this is a subjective design decision rather than anything more concrete. (see attached figure for an example)
2. Currently it looks like there may be some leaked memory around PortamentoRealtime. The easy way to see this is to switch the realtime memory allocator into a normal libc malloc/free combination. Around src/Misc/Allocator.cpp:77..86 just swap the tlsf_malloc and tlsf_free lines with the malloc/free lines commented out. Then you can use tools like valgrind to check for leaked memory. In this case valgrind --leak-check=full ./build/src/Tests/KitTest has the attached output. I haven't checked if this is only an issue within the unit tests, but it would be ideal if leaks were avoided there.
3. I almost missed this on the first pass, but PortamentoRealtime has a callback function. Lambdas are excellent and I absolutely love that C++ has them, but one implementation detail is that lambdas with captures will allocate memory to store captured values on the heap. So, src/Misc/Part.cpp:578 allocates memory which is ideally avoided on the realtime side. The usual way to work around the problem is to have something like a void* argument which is passed back into a capture-free lambda. Annoying to do, but it's the only workaround that I know of.
4. The feature is great, so I fully support throwing one of your previously written descriptions into the doc/ directory. It's not heavily used, but it's more likely to survive over time than github issues which IMO are tricky to search through.


```
==27599== 
==27599== HEAP SUMMARY:
==27599==     in use at exit: 119,216 bytes in 1,364 blocks
==27599==   total heap usage: 149,332 allocs, 147,968 frees, 55,400,421 bytes allocated
==27599== 
==27599== 64 bytes in 1 blocks are definitely lost in loss record 45 of 96
==27599==    at 0x483E77F: malloc (vg_replace_malloc.c:309)
==27599==    by 0x46CAD8: zyn::AllocatorClass::alloc_mem(unsigned long) (Allocator.cpp:78)
==27599==    by 0x458AE4: zyn::PortamentoRealtime* zyn::Allocator::alloc<zyn::PortamentoRealtime, zyn::Allocator&, zyn::Part::NoteOnInternal(unsigned char, unsigned char, float)::{lambda(zyn::PortamentoRealtime*)#1}, zyn::Portamento&>(zyn::Allocator&, zyn::Part::NoteOnInternal(unsigned char, unsigned char, float)::{lambda(zyn::PortamentoRealtime*)#1}&&, zyn::Portamento&) (Allocator.h:42)
==27599==    by 0x44F43A: zyn::Part::NoteOnInternal(unsigned char, unsigned char, float) (Part.cpp:574)
==27599==    by 0x43448B: zyn::Part::NoteOn(unsigned char, unsigned char, int) (Part.h:53)
==27599==    by 0x4387F2: KitTest::testPortamentoOnPlayingLegato() (KitTest.cpp:1095)
==27599==    by 0x434133: main (KitTest.cpp:1515)
==27599== 
==27599== 64 bytes in 1 blocks are definitely lost in loss record 46 of 96
==27599==    at 0x483E77F: malloc (vg_replace_malloc.c:309)
==27599==    by 0x46CAD8: zyn::AllocatorClass::alloc_mem(unsigned long) (Allocator.cpp:78)
==27599==    by 0x458AE4: zyn::PortamentoRealtime* zyn::Allocator::alloc<zyn::PortamentoRealtime, zyn::Allocator&, zyn::Part::NoteOnInternal(unsigned char, unsigned char, float)::{lambda(zyn::PortamentoRealtime*)#1}, zyn::Portamento&>(zyn::Allocator&, zyn::Part::NoteOnInternal(unsigned char, unsigned char, float)::{lambda(zyn::PortamentoRealtime*)#1}&&, zyn::Portamento&) (Allocator.h:42)
==27599==    by 0x44F43A: zyn::Part::NoteOnInternal(unsigned char, unsigned char, float) (Part.cpp:574)
==27599==    by 0x43448B: zyn::Part::NoteOn(unsigned char, unsigned char, int) (Part.h:53)
==27599==    by 0x438814: KitTest::testPortamentoOnPlayingLegato() (KitTest.cpp:1096)
==27599==    by 0x434133: main (KitTest.cpp:1515)
==27599== 
==27599== 64 bytes in 1 blocks are definitely lost in loss record 47 of 96
==27599==    at 0x483E77F: malloc (vg_replace_malloc.c:309)
==27599==    by 0x46CAD8: zyn::AllocatorClass::alloc_mem(unsigned long) (Allocator.cpp:78)
==27599==    by 0x458AE4: zyn::PortamentoRealtime* zyn::Allocator::alloc<zyn::PortamentoRealtime, zyn::Allocator&, zyn::Part::NoteOnInternal(unsigned char, unsigned char, float)::{lambda(zyn::PortamentoRealtime*)#1}, zyn::Portamento&>(zyn::Allocator&, zyn::Part::NoteOnInternal(unsigned char, unsigned char, float)::{lambda(zyn::PortamentoRealtime*)#1}&&, zyn::Portamento&) (Allocator.h:42)
==27599==    by 0x44F43A: zyn::Part::NoteOnInternal(unsigned char, unsigned char, float) (Part.cpp:574)
==27599==    by 0x43448B: zyn::Part::NoteOn(unsigned char, unsigned char, int) (Part.h:53)
==27599==    by 0x438836: KitTest::testPortamentoOnPlayingLegato() (KitTest.cpp:1097)
==27599==    by 0x434133: main (KitTest.cpp:1515)
==27599== 
...
```



2021-11-06-portamento



Thanks a ton for the effort of detangling things and the very clean code to boot. Once comments 2/3 are addressed this should be immediately rebased&merged.

polluxsynth:
Thanks for your comments!

If I understand 1. correctly, you are playing non-legato (i.e. releasing every note before the next one is played) consistently, yet sometimes you get portamento when playing a lower note? If so that's not intentional, in mono/legato mode there should only be portamento when playing legato, although that includes the following scenario: first playing one note, holding it and playing a second one, then still holding the first, releasing the second. In this case there should be portamento gliding up to the second note, and then down to the first one when the second one is released, by the logic that one is one is still playing legato when holding the first note and releasing the second (envelopes don't retrigger for instance).

Regarding 2. I suspect it's only in the test because I don't clean things up properly in the tests, but I agree that the tests should not leak memory even if it's not as detrimental as the main application doing so, so I'll have a look at this. Thanks for the instructions on setting it up with standard malloc/free and valgrind.

Regarding 3. yeah,. that's a bummer. I'll have to rework this.

And finally, regarding 4, should I include a doc/ update in the PR then?

fundamental:
Per 1 I'm running the arpeggiator on my MIDI keyboard with a high BPM. Gate is below 100% so, it should note on(a), wait, note off(a), wait, note on(b), ... Zyn is at default settings with the key mode changed to either legato or mono (both behave the same in this test).

> first playing one note, holding it and playing a second one, then still holding the first, releasing the second. In this case there should be portamento gliding up to the second note, and then down to the first one when the second one is released, by the logic that one is one is still playing legato when holding the first note and releasing the second (envelopes don't retrigger for instance).

I can confirm that scenario is working correctly and is a load of fun to play around with. Usually I just hold one note and tap a second one to juggle the portamento up and down. Great feature and not the curious behavior originally described.

Per 4, Sure just add an addition commit leaving a .txt file in that folder. All the other docs there should be asciidoc which is pretty darn close to the markdown your comment is already written in.

polluxsynth:
Beg pardon, but are you sure your MIDI keyboard is actually outputting what you think it's outputting? I.e. have you verified that the MIDI data which is sent actually has the note off before the note on every time?

The only MIDI keyboard I have which can send arpeggiated notes over MIDI is the NordLead 2X. It doesn't have a gate setting, but it sounds like it's trying to be 100% (i.e. no time between note off and subsequent note on, no matter how slow I run it). Using the 2X arpeggiator, no matter how fast I run it, I don't get portamento (whereas I do get it when playing legato manually).

I also tried with the sequencer in my E-Mu XL-7. Setting the gate to 100% when recording a sequence results in no portamento, the same actually with the gate set to 101%, which is probably a bug in its firmware, but 102% results in consistent portamento.

I haven't checked exactly what the MIDI timing is on the data sent by these two machines, but at least they do seem to be consistent with what I expect (except for the 101% gate setting on the XL-7).

polluxsynth:
Good catch on the memory leak, I had missed the case of notes being killed, so the problem was real, not just in the test.

I noticed however that valgrind complains a lot about:

==9125== Conditional jump or move depends on uninitialised value(s)
==9125==    at 0x1C3572: zyn::AnalogFilter::setfreq_and_q(float, float) (in /home/ricard/src/zyn-fusion-build/build/build-zynaddsubfx-linux-release/src/Tests/KitTest)

I'm assuming that's to be expected (for whatever reason, have not looked into it)?

fundamental:
Good catch. I'll commit this a bit later as it's unrelated to these changes, but the setfreq_and_q() issue is resolved via:

```
diff --git a/src/DSP/AnalogFilter.cpp b/src/DSP/AnalogFilter.cpp
index ef9dbaf0..036c0770 100644
--- a/src/DSP/AnalogFilter.cpp
+++ b/src/DSP/AnalogFilter.cpp
@@ -44,6 +44,7 @@ AnalogFilter::AnalogFilter(unsigned char Ftype,
     if(stages >= MAX_FILTER_STAGES)
         stages = MAX_FILTER_STAGES;
     cleanup();
+    beforeFirstTick=true;
     setfreq_and_q(Ffreq, Fq);
     coeff.d[0] = 0; //this is not used
     outgain    = 1.0f;
@@ -292,7 +293,7 @@ void AnalogFilter::setfreq(float frequency)
         freq = frequency;
         recompute = true;
     }
-    
+
     if (beforeFirstTick) {
         freq_smoothing.reset( freq );
         beforeFirstTick=false;
```

Now to see about verifying the note sequencing.

fundamental:
Looks like everything is being emitted from my keyboard. I've attached an event log of note on/note off/audio out. Not a full replication, but it does look like my MIDI device has the correct behavior. If you want a perfect 'this exact input triggers the bug' I can run another test.

[midi-event-log.txt](https://github.com/zynaddsubfx/zynaddsubfx/files/7492373/midi-event-log.txt)

polluxsynth:
Looking at the event log, I see several cases of the following type of sequence:

`note on <chan:0 note:106 vel:100>`
...
`note on <chan:0 note:104 vel:100>`
`note on <chan:0 note:106 vel:0>`
`note off <chan:0 note:106>`

(the two last lines unsurprisingly appear to refer to the same event  (because a note on with velocity 0 is considered a note off), but it is clear that from time to time the note on for a note is received before the note off for a previous note.

In many cases it does come out right, i.e. 

`note on <chan:0 note:104 vel:0>`
`note off <chan:0 note:104>`
`note on <chan:0 note:106 vel:100>`

So it still seems to me that the MIDI keyboard is outputting the events in such an order that they would be considered legato.

fundamental:
Hm, I guess that might be a defect in our ALSA code. I wouldn't expect anything inconsistent out of the keyboard as it's a reputable brand (Arturia Keystep), but I wouldn't be surprised if there's something quirky with event handling in src/Nio/AlsaEngine.cpp . I'll take a stab this week with jack MIDI input which should avoid any potential ambiguities.

polluxsynth:
Honestly, I've seen strange behavior from reputable devices before. For the record, I was using ALSA MIDI too when testing. But of course the NordLead 2X and E-Mu XL-7 that I was testing with might have some slight difference in MIDI timing compared to the Keystep which would account for the difference.

Is there any way you can record the output from your Keystep into a sequencer of some sort, or indeed any MIDI capture software or device?

polluxsynth:
A friend of mine has a Keystep that I can borrow for a couple of days to make further investigations.

polluxsynth:
So I borrowed my friend's Keystep but try as I might, I can't reproduce any occasional portamento behavior with the arpeggiator switched on. With a Zyn patch set to legato mode, with portamento enabled, the pitch glides nicely when playing legato (manually), and with the arpeggiator running there's no legato even with the rate set to maximum, and 1/32 or 1/32T time division. Tried both with the Keystep being connected via USB to my computer and via a DIN MIDI using a separate USB MIDI interface.

So this is a weird one. The thing I'm wondering about at the moment is if your Keystep has the latest firmware, although I can't see anything in the bug list on Arturia's web site that indicates that earlier versions had this particular problem.

fundamental:
Hm. Now I'm really quite confounded. I'll replicate tomorrow morning, but the general steps were. Start jackd in dummy mode, start zyn with alsa input+jack output, connect zyn to baudline, connect keystep via aconnect, set key mode to legato or mono using oscprompt, shift the keystep up to max octave shift (makes it easier to read the spectrogram), press and hold several keys with rate set to a high value and time division set to one of the fast settings. Keymode was one of the up-down modes as can be seen from the screenshot.

You not seeing the problem with the same MIDI hardware makes me think that I've done some sort of mistake on my end, but I'll have to see if I fail to replicate the issue tomorrow. The biggest difference in setup is likely the notes that were chosen for input as shifting up the few octaves may have been a needed step.

I'll have to take a look into device firmware. I never updated mine and I'm not even sure if such a utility exists on Linux, though I haven't checked.

polluxsynth:
I honestly didn't try with the octave shift set to more than two, I'll give it a try maximizing it, and see what happens.

fundamental:
I can confirm that it's a mistake on my end. The Arp/Seq switch doesn't stay in position as well as I'd like, so I had not noticed when it moved. Testing in arp mode even on high gate levels I cannot replicate the bug, so the logic is all fine. My bad for raising a false alarm.

polluxsynth:
Ok, no problem, good to hear that it's working fine at your end. I've fixed issues 2 and 3, so I'll have a look at 4 within a couple of days hopefully and then submit a new PR.

zynmuse:
Thanks polluxsynth for running with this.
I can see  it's been well thought through.
It will be good to have your improvements implemented.

polluxsynth:
How are the files in the doc/ directory actually used? I'm planning to add the portamento info to controller.txt, but when testing by touching the file and doing `make -C  build/build-zynaddsubfx-linux-release` or `make -C  build/build-zynaddsubfx-linux-release doc` nothing is rebuilt, and if I go into src/zynaddsubfx/doc and type `make` I get dumped into pdflatex which claims that `! LaTeX Error: File 'standalone.cls' not found.`, asking for which file to use instead.

polluxsynth:
So finally here is the updated PR. As noted above, I've addressed issues 2 and 3, and addressed issue 4 by adding the portamento description texts from issue #304 in revised form to doc/controller.txt, with slight modifications of the existing text.

fundamental:
Just updating. An unusual crunch period with my day job has delayed zyn work by a few days, but I'm putting time aside Wednesday evening to address this and instruments pr 4.

fundamental:
Excellent. Thanks for the well built contribution :)

polluxsynth:
Thanks Mark!

Next step is to implement 'auto glide' mode, but will take a couple of weeks before I have time to work on it.

PR #152, opened 2021-10-31
Add an option to include disabled parts in the XML save file (again)

zynmuse:
This ensures all parts and kits that contain changed parameters are now saved to XML regardless of whether they are enabled.
Now makes use of xml.minimal and also adds a new parameter to SaveFullXml to xml class.


fundamental:
Unit tests are still failing and the commit sequence is getting pretty messy.

PR #154, opened 2021-11-18
Fix naming of jack_client using NSM client-id when using JACK-MULTI o…

orlammd:
…utput


fundamental:

PR #157, opened 2021-12-18
FFTwrapper: Use float precision

JohannesLorenz:


falkTX:
Is this happening?

I remember I did a patch quite some years ago for this.
Might be useful as reference/comparison https://github.com/moddevices/mod-plugin-builder/blob/master/plugins/package/zynaddsubfx-labs/08_use-fftwf.patch

fundamental:
It's quite likely happening. I would have sworn I tried to merge in the relevant bits from the mod patch set in the past, but I might have overlooked one. Looks like from the current patch set, this is relevant and perhaps the no_asm patch would be good to merge after a quick sanity check.

falkTX:
That would be cool to see.
The float version is typically less CPU intensive.

I recommend to change the `fftw_real` name to `fftwf_real` too, as I have done in the MOD patch.
That makes it clear that the type is intentionally a float, because `fftw_` prefixed calls and types usually refer to double.

fundamental:
I'd agree with the fftw_real -> fftwf_real. Otherwise looks good to go.

@falkTX know of any platforms where distributions would build fftw without single precision mode? I'm not even sure if that's a flag with current releases, though I seem to remember there being some sort of oddball hangup with the fftwf_() interface a decade ago.

JohannesLorenz:
Fixed `fftwf_real` and converted more doubles according to the reference.

`git grep double` still shows some more doubles (outside the fft stuff). Should I elliminate them, too?
```
src/Nio/AlsaEngine.cpp:    double scaled;
src/Nio/SndioEngine.cpp:    double scaled;
src/Synth/OscilGen.cpp:power_cosinus_32(float _x, double _power)
src/Synth/OscilGen.cpp:    double retval;
src/Synth/OscilGen.cpp:power_sinus_32(float _x, double _power)
```

The OscilGen doubles seem to be non-RT. The Engine doubles can be fixed IMO. Agreed?

fundamental:
doubles aren't intrinsically evil, so I wouldn't be all that worried by their existence elsewhere, only in performance heavy blocks.

JohannesLorenz:
By the way: I just did some benchmark tests on a "usual" desktop PC (`Intel(R) Core(TM) i5-6600K CPU @ 3.50GHz`, 4 cores, ~5 years old), compiled with `CMAKE_BUILD_TYPE=Release`. I modified PadNoteTest to have maximum octaves, smps/octave and samplesize, in order to get a really long computation.

```
Before this PR: 6.21s
After this PR: 3.51s
```

falkTX:
> I'd agree with the fftw_real -> fftwf_real. Otherwise looks good to go.
> 
> @falkTX know of any platforms where distributions would build fftw without single precision mode? I'm not even sure if that's a flag with current releases, though I seem to remember there being some sort of oddball hangup with the fftwf_() interface a decade ago.

I have never seen a single problem with fftwf usage. It is still behind --single build flag afaik, but distros take it to be important enough for packaging.

JohannesLorenz:
Also, it works in LMMS, and LMMS runs on pretty much every platform. Their FindFFTW.cmake is a bit more complicated than ours, but I'm for keeping it simple as long as no one complains.

Let's merge this PR?

fundamental:
Merged :+1:

PR #158, opened 2021-12-21
src/CMakeLists.txt: use the target compiler during config tests

jolivain:
The availability of the "fistpl" x86 instruction was tested with a
direct call to the "as" command. In case the project is cross-compiled
to another architecture, the host compiler is tested instead the
target one. This was leading to compilation failures. This commit now
perform the test using the target compiler defined in CMake, which
fixes cross-compilation. This commit does not change the behavior
when doing a native compilation, as the host and target compiler are the
same.

Signed-off-by: Julien Olivain <ju.o@free.fr>

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -59,9 +59,10 @@ find_library(HAVE_LIBRT rt)
 set(CMAKE_REQUIRED_FLAGS "")
 
 
-execute_process(COMMAND echo fistpl 0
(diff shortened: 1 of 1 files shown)


fundamental:
I checked the first commit a few times as I've made incorrect assumptions about how cmake interprets the results of ASM tests before, but this addition appears to work as intended.

Thanks for the enhancements :+1:

PR #159, opened 2021-12-31
Fix gcc warnings

jolivain:
This pull request fixes few compiler warnings reported by gcc 11 on Linux.
Those changes are mostly cosmetic and are not changing the behavior.


fundamental:
Looks good to me. Thanks for fixing these issues :+1:

PR #161, opened 2022-01-02
Removing buffers from FFTwrapper

JohannesLorenz:
Reason for PR: parallelize `FFTwrapper` => remove its buffers.
Solution: run directly on input/output buffers => ensure that these are big enough, using new classes `FFTsampleBuffer` and `FFTfreqBuffer`.
Alternative solution (not realized): pass temporary buffers to `FFTwrapper` for every FFT/IFFT (probably more complicated, would take more memcpys)


fundamental:
Looks good and I think we discussed any of the issues with the previous version. As tests are passing I'd say all's ready to merge :+1:

PR #162, opened 2022-01-02
OscilGen: Make sure fftsize matches oscilsize

JohannesLorenz:
This was required to make sure that the OscilGen matches the FFTwrapper.
This also covers an issue in UnisonTest which is fixed in this commit.


fundamental:
Merged.

PR #163, opened 2022-01-04
Fix all tests to pass valgrind's leak checks

### LINKED ISSUE #160: Potential memory leak in PresetExtractor
Code:

```
if(xml.enterbranch(type) == 0)
        return;
```

This does not delete the previously allocated `t`.

JohannesLorenz:
Now ready to review.
Should not be anything new after the IRC discussions.

Fixes #160

### DIFF
--- a/src/Misc/PresetExtractor.cpp
+++ b/src/Misc/PresetExtractor.cpp
@@ -227,7 +227,10 @@ void doPaste(MiddleWare &mw, string url, string type, XMLwrapper &xml, Ts&&... a
         type = "Plfo";
 
     if(xml.enterbranch(type) == 0)
+    {
+        delete t;
         return;
+    }
 
     t->getfromXML(xml);
 
--- a/src/Tests/MiddlewareTest.cpp
+++ b/src/Tests/MiddlewareTest.cpp
@@ -33,9 +33,10 @@ char *instance_name=(char*)"";
 
 #define NUM_MIDDLEWARE 3
 
-class PluginTest
+class MiddleWareTest
 {
     public:
+        struct FFTCleaner { ~FFTCleaner() { FFT_cleanup(); } } cleaner;
         Config config;
         void setUp() {
             synth = new SYNTH_T;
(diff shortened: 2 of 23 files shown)


src/Misc/PresetExtractor.cpp
```diff
@@ -227,7 +227,10 @@ void doPaste(MiddleWare &mw, string url, string type, XMLwrapper &xml, Ts&&... a
         type = "Plfo";
 
     if(xml.enterbranch(type) == 0)
+    {
+        delete t;
```
fundamental, line 231: Same comment applies to the old code, but should we consider this a bug if we ever get to this return statement?

src/Tests/MiddlewareTest.cpp
```diff
@@ -60,16 +61,21 @@ class PluginTest
                 middleware[i] = new MiddleWare(std::move(*synth), &config);
                 master[i] = middleware[i]->spawnMaster();
                 //printf("Octave size = %d\n", master[i]->microtonal.getoctavesize());
+                if (i != NUM_MIDDLEWARE-1) {
+                    delete synth;
```
fundamental, line 65: This is required even with the std::move? I tend to forget some of those semantics.

fundamental:
Perhaps a few extra free methods for the various zyn types, but no harm other than a few extra ms during the build phase. Just confirm that the std::move(*synth) delete looks right to you on a second glance I'm I'd say let's merge this.

src/Tests/MiddlewareTest.cpp
JohannesLorenz: [Moving an object does not change its lifetime](https://stackoverflow.com/questions/15663539/why-isnt-object-returned-by-stdmove-destroyed-immediately). Compare it with an `std::vector` that you have on the stack - the move only alters its pointers. After you move from it, the memory is still there. Same with a vector (or `SYNTH_T`) on heap.

Aside from the theory, I could see with valgrind that this fixed a leak.

src/Misc/PresetExtractor.cpp
JohannesLorenz: I think so, but what is our way to handle bugs? Should zyn exit just because of a failed paste?

Another thing we could do is print a message on `return`ing, so that the problem is revealed.

src/Tests/MiddlewareTest.cpp
fundamental: Works for me, merging now.

PR #164, opened 2022-01-05
Jack multi part mute

polluxsynth:
This turned out to be slightly more complex than I anticipated, and also there already seemed to be code which disables a part gracefully which has not actually been used for a while.


fundamental:
Clear explanation in the commits and it matches the described behavior. Thanks for the enhancement :+1:

PR #165, opened 2022-01-07
Parallelize oscilgen

JohannesLorenz:


fundamental:
I was trying to figure this out yesterday, but any idea why the CI is trying to run this pull request as a push and a pull request? I'm guessing it's a typo on my end, but given that searching for the error didn't yield any results there's a slim chance it might be a github bug.

JohannesLorenz:
I think I did a force push while the first build was running, and that confused github. Should I close this and submit a new PR?

src/Synth/OscilGen.cpp
```diff
@@ -38,9 +38,9 @@ const rtosc::Ports OscilGen::non_realtime_ports = {
     rPaste,
     //TODO ensure min/max
     rOption(Phmagtype, rShort("scale"),
-            rOptions(linear,dB scale (-40),
-                     dB scale (-60), dB scale (-80),
-                     dB scale (-100)),
+    rOptions(linear,dB scale (-40),
```
JohannesLorenz, line 41: This change here was just dumbness... :see_no_evil:

fundamental:
No need to resubmit, the pull_request CI worked fine. The push CI seems to be searching for a commit that it can't find, but if it's an out-of-date commit anyhow it doesn't matter.

fundamental:
Reading through the changes I'm not spotting anything which looks like it's out of place. I was going to say 'mybuffers' was a bit uncreative, but the second commit nukes any references to that variable name. Right now everything seems right, but I'm depending in large parts on the unit tests and their not-excellent coverage to see if anything is semantically different. I think it's good to merge now.

src/Synth/OscilGen.cpp
JohannesLorenz: Fixed in a further commit.

JohannesLorenz:
I pushed a fixup. CI passed now

The second commit did not nuke all `myBuffers`, though there is a chance that this variable may vanish completely in the future. If you have better names, let me know.

fundamental:
Eh, don't worry about the naming too much as it's generally moved towards a more concise version in the newer commits in the PR. Thanks for fixing up parts of oscilgen :)

PR #166, opened 2022-01-08
Make tests independent of config

JohannesLorenz:
This uses a fixed config in tests, instead of relying on the user
config.

For `MiddleWare::saveParams`, the second master (which controls the
saving) must have the same config values for `SaveFullXml` as the first
master (the original one). Otherwise, they dump different XMLs which can
not be compared.


fundamental:
Looks good to me and I can confirm that the global config is only altered when config.save() is called in exitprogram() in main.cpp (sorry plugin users). So as soon as the CI says it's good I'll merge it in.

PR #169, opened 2022-01-18
Misc/Part: Fix LR swap in multi mode (#330)

polluxsynth:
Fix inverted LR panning in part outputs in multi mode. This was due
to LR being swapped in three places, in conjunction with the part
panning control being inverted:
- Mixer VU meter LR was swapped
- Part VU meters LR were swapped
- Audio mix output LR was swapped

A partial clue to the above (apart from the obviously inverted
LR part panning) was that panning in the parts effects (e.g. Echo)
was audibly ok in multi mode, but the VU meter indication was
swapped Left-Right. At the same time, the summed output from the
mixer exhibited inverted panning with regards to panning set in a
part effect.

### DIFF
--- a/src/Misc/Master.cpp
+++ b/src/Misc/Master.cpp
@@ -1057,7 +1057,7 @@ void Master::setController(char chan, int type, note_t note, float value)
             part[npart]->SetController(type, note, value, keyshift);
 }
 
-void Master::vuUpdate(const float *outr, const float *outl)
+void Master::vuUpdate(const float *outl, const float *outr)
 {
     //Peak computation (for vumeters)
     vu.outpeakl = 1e-12;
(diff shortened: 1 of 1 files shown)


fundamental:
Looks to be consistent with the preexisting header files (unlike the old code), so I'm good with merging. Worst case for the non-multi case some people may need to toggle the lr swap global config.

Thanks for the fix :+1:

polluxsynth:
Thanks for merging!

I tested the non-multi case both before and after the fix, and the inverted part panning and audio mix bus cancel each other out in terms of part panning.

However, it is true that if something in a patch itself was panned left it will now be panned right, or rather, if it was panned left before then it would have appeared to be right in the summed mix, but it is now panned left as it should have been.

fundamental:
A possible fix is to check the old file version numbers and to flip panning, though in this case I think I'll just add it to the changelogs as a note.

polluxsynth:
You mean check the version numbers in a patch file when it is loaded, and behave accordingly?

fundamental:
> You mean check the version numbers in a patch file when it is loaded, and behave accordingly?

Almost. Basically the logic in other places is:

if(is_older(file_version, target_version))
    parameter_x = remap(load_parameter('x'))
else
    parameter_x = load_parameter('x')

polluxsynth:
This sounds like something which would solve the problem I've got with implementing Auto portamento mode, where the Auto setting for old patches would have to vary depending on which key assign mode was used. (I think you said that it didn't really matter if the behavior changed in this respect, but it's still been bugging me).

fundamental:
Yep, it should help in any of the situations where behavior changes.

PR #170, opened 2022-01-22
Add cmake option to optionally use system rtosc

dvzrv:
CMakeLists.txt:
Add `ZYN_SYSTEM_RTOSC` option (defaults to OFF) to optionally use
pkgconf integration to find and use a system provided librtosc.

Related to https://github.com/zynaddsubfx/zyn-fusion-issues/issues/216


fundamental:
I don't think this works based on my first read through, but it should only require minor alterations. The CMake build process uses the colors defined in rtosc to generate out the configuration summary.

I think if the color cmake file cannot be loaded from rtosc, then zyn needs to define the escape codes for:

- Green
- Red
- Yellow
- ColorReset

Those get used in src/CMakeLists.txt:661..692

fundamental:
Actually I take that back, cmake just allows things to be undefined (great for debugging), so the missing escape codes are canceled out by the lack of clearing codes. Time to give this another look and just merge acknowledging that there won't be colors in an external rtosc build.

fundamental:
Yep, looks good, merging :+1:

PR #171, opened 2022-01-22
Allow linking against system rtosc and rtosc-cpp

dvzrv:
CMakeLists.txt:
Also check for librtosc-cpp when `ZYN_SYSTEM_RTOSC=ON`.

src/CMakeLists.txt:
When `ZYN_SYSTEM_RTOSC=ON`, link against system provided librtosc and
librtosc-cpp, else link against the local targets (exposed via the rtosc
submodule).

This is a follow-up to https://github.com/zynaddsubfx/zynaddsubfx/pull/170 to ensure proper linking against system provided libraries.


fundamental:
Sounds like this one has a better chance of success than the last one :)

PR #172, opened 2022-01-22
Add CI integration for Arch Linux

dvzrv:
.github/workflows/ccpp.yml:
Build in an Arch Linux container against system rtosc and install the
resulting software.


fundamental:
Looks good to me :+1:

PR #174, opened 2022-01-24
use of NSL naming for alsa port as well

orlammd:
in NSM context only.


fundamental:
Thanks for the enhancement :+1: I'm using squash+merge here to eliminate the duplicate commits.

PR #175, opened 2022-01-29
Auto portamento mode

polluxsynth:
Implement auto mode for portamento. This decouples the requirement to play legato to get portamento for the mono and legato modes from the mode itself, and allows activation of portamento in poly mode to be optionally controlled by legato playing, governed by a new portamento parameter.

I've let the automode member for the Controller.portamento struct be an unsigned char as all others are unsigned chars, and it didn't seem right to attempt to clean this up at this time.

In order to set auto portamento mode selectively when loading older patches, based on the key assign mode, I've bumped the revision version number. I don't know if this is an ok thing to do at this point.

### DIFF
--- a/src/Params/Controller.cpp
+++ b/src/Params/Controller.cpp
@@ -62,6 +62,8 @@ const rtosc::Ports Controller::ports = {
         "Portamento MIDI Receive"),
     rToggle(portamento.portamento, rDefault(false),
         "Portamento Enable"),
+    rToggle(portamento.automode, rDefault(false),
+        "Portamento Auto mode"),
     rParamZyn(portamento.time,          rShort("time"), rDefault(64),
         "Portamento Length"),
     rToggle(portamento.proportional,    rShort("propt."), rDefault(false),
@@ -112,6 +114,7 @@ void Controller::defaults()
     NRPN.receive                 = 1;
 
     portamento.portamento        = 0;
+    portamento.automode          = 0;
     portamento.proportional      = 0;
     portamento.propRate          = 80;
     portamento.propDepth         = 90;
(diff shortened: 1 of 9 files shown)


src/Params/Controller.cpp
```diff
@@ -112,6 +114,7 @@ void Controller::defaults()
     NRPN.receive = 1;
 
     portamento.portamento = 0;
+    portamento.automode = 0;
```
fundamental, line 117: Very minor note, so I wouldn't worry about fixing this on this PR, but the rest of the initializers appear to be vertically aligned.

fundamental:
The code makes sense and I'm 100% fine with this introducing the version upgrade to ensure comparability with older settings. I still have to try out the implementation as an additional sanity check, but I'd assume that will be a smooth process.

src/Params/Controller.cpp
polluxsynth: It's easy enough to fix though, although honestly it looks like the whole section could do with a cleanup, with some initializers being aligned and some not. It's no problem for me to do an updated PR.

polluxsynth: I cleaned up the defaults() initializations as a separate commit. The rest of the commits are the same.

src/Params/Controller.cpp
```diff
@@ -112,6 +114,7 @@ void Controller::defaults()
     NRPN.receive                 = 1;
 
     portamento.portamento        = 0;
+    portamento.automode          = 0;
```
fundamental, line 117: After trying out the code I think that setting automode to 1 by default makes the most sense. Any opposition to that change?

polluxsynth: Not much, it really is your call. My rationale was that automode modifies the basic portamento function and makes the behavior more intricate, and as such should be disabled by default as that reflects the simplest behavior. On the other hand, a rationale for the opposite would be that portamento is most likely used in the monophonic modes, where automode previously was the only behavior, and so for reasons of familiarity should be enabled by default.

fundamental:
Thanks for the enhancement. I added in the auto mode default change as a subsequent commit :+1:

PR #176, opened 2022-01-31
Use of lo_send_message_from in sendToRemote method

orlammd:


src/Misc/MiddleWare.h
```diff
@@ -13,6 +13,7 @@
 #include <functional>
 #include <cstdarg>
 #include <string>
+#include <lo/lo.h>
```
fundamental, line 16: lo/lo.h is included within MiddleWare.cpp and is not required in MiddleWare.h

fundamental:
Looks like there's the one deletion to consider and ideally the PR should be a single commit. Looking at this webpage you can see that it includes 6 commits. To avoid this you'd typically start with a checkout at the zynaddsubfx/zynaddsubfx git master HEAD, then create a branch which you can push to your fork. The fork would then have a branch at the same commit id as zynaddsubfx/zynaddsubfx's master branch. From there you'd make the changes in the PR, create the pull request (request your branch in your fork gets merged with zynaddsubfx/zynaddsubfx's master branch), and then there would only be the desired commits. After the PR is merged then typically the branch would be deleted on your fork.

That way you can have a forked repo which would have an identical master branch, so multiple pull requests don't drift apart. Hopefully that makes sense as github doesn't really provide many hints along these lines.

falkTX:
we can just squash all the commits into 1 when merging, no?

orlammd:
Hi!

Le 02/02/2022 à 04:19, Mark McCurry a écrit :
> Looks like there's the one deletion to consider and ideally the PR 
> should be a single commit. Looking at this webpage you can see that it 
> includes 6 commits. 

Yes, sorry, I'm not a git master at all, and only know the minimum 
requirements about it!
Moreover the fact is the git repo I work on my laptop is a submodule of 
zyn-fusion, and I got problems merging with the fork I made on 
PlagiatBros, so I might have not chosen the good solutions...

> To avoid this you'd typically start with a checkout 
> at the zynaddsubfx/zynaddsubfx git master HEAD, then create a branch 
> which you can push to your fork. The fork would then have a branch at 
> the same commit id as zynaddsubfx/zynaddsubfx's master branch. From 
> there you'd make the changes in the PR, create the pull request (request 
> your branch in your fork gets merged with zynaddsubfx/zynaddsubfx's 
> master branch), and then there would only be the desired commits. After 
> the PR is merged then typically the branch would be deleted on your fork.
> 
> That way you can have a forked repo which would have an identical master 
> branch, so multiple pull requests don't drift apart. Hopefully that 
> makes sense as github doesn't really provide many hints along these lines.

You want me to do that for this PR, or that's an advice for the 
potential next(s) to come?

fundamental:
> You want me to do that for this PR, or that's an advice for the
potential next(s) to come?

No need to do it for this one if you're fine with a squash and merge. The change to MiddleWare.h needs to be resolved, but the comment is mostly for future PRs if you have other pending changes you're interested in making.


> we can just squash all the commits into 1 when merging, no?

Yep, though I usually like to stick to the rebase+merge when possible. Keeps any changes from my end to a minimum which makes it easier to review changes when releases crop up (e.g. the author attribution script for commit counts). Just seems cleaner unless there's more of a reason to intentionally want to squash. i.e. there's a lot of little changes that only end up making sense as an atomic commit, but when the patchset was created it made sense to keep things more granular.

jean-emmanuel:
I've updated the PR to keep only the relevant commit.

fundamental:
Merged :+1:

PR #181, opened 2022-03-03
Improve bank ops

polluxsynth:
These are some fixes needed to support the bank read and write improvements in Fusion.


src/Misc/MiddleWare.cpp
```diff
@@ -1535,13 +1537,14 @@ static rtosc::Ports middwareSnoopPortsWithoutNonRtParams = {
         const int slot    = rtosc_argument(msg, 1).i;
 
         int err = 0;
+        printf("save_to_slot: part %d slot %d, filename %s\n", part_id, slot, impl.master->part[part_id]->Pname);
```
fundamental, line 1540: I'm guessing these printfs were unintentionally left in place.

fundamental:
Everything looks reasonable reading through the code, now all that's left is to run both PRs and see if anything unexpected happens at runtime.

src/Misc/MiddleWare.cpp
polluxsynth: Yes, good catch, this one (line 1540) and the one on line 1543. Do you want to resubmit the PR with those removed (probably best to submit a new one rebased to the latest master)?

src/Misc/Part.cpp
```diff
@@ -179,20 +179,24 @@ static const Ports partPorts = {
...
             else
             {
-                p->saveXML(p->loaded_file);
-                fprintf(stderr, "Part %d saved to %s\n", (p->partno + 1), p->loaded_file);
+                err = p->saveXML(p->loaded_file);
+                fprintf(stderr, "Part %d saved to %s: %s\n", (p->partno + 1), p->loaded_file, err ? "failed" : "ok");
             }
+            if (err)
+                d.reply("/alert", "s",
+                        "Failed To Write, please check file permissions");
```
hselasky, line 199: s/Failed To Write/Failed to save file/

???

polluxsynth: Yes, it makes sense, on the other hand, the function is mostly used when clicking OVERWRITE in READ mode in Fusion, so I chose 'Failed To Write' as a natural response to 'Overwrite'. But it would also be used when doing Save Instrument from the File menu (which however when testing right now doesn't seem to emit the /alert for whatever reason), which case 'Failed to save file' would be more appropriate, and considering the name of the OSC call (savexml) your suggestion also makes more sense.

I've got a couple of changes lined up for this PR based on comments from @fundamental so I'll add this change to that and send them all off together with any other changes that Mark might have regarding the functionality itself.

polluxsynth:
@fundamental , I've adjusted the patch set based on your and @hselasky 's comments above, but I thought I'd hold off on pushing it until you've done some runtime testing in case there are other things that need fixing. Or would you rather that I send what I have now as soon as possible?

fundamental:
I'll target doing the runtime eval this afternoon. If something ends up blocking me from doing so by the end-of-day, feel free to add in the other changes.

polluxsynth:
Removed the unintentionally forgotten printfs, and changed the alert message as per @hselasky 's comments, rebased to latest master and repushed. Not sure if the rebase has got github confused; it would seem so when clicking the latest Compare button above (because it compares with the previous push), but the individual commits are available so I'm assuming it's not a problem, let me know otherwise.

fundamental:
I suspect the code within this PR is correct, though it does make it possible to trigger a bug that I've been looking to localize. Triggering a bank write to a slot which cannot be written to due to file permissions will consistently produce `Unknown address<BACKEND:online>` messages with multiple response messages which should be sent to the GUI getting redirected to the backend. Somehow the d.reply("/alert", ...) code is triggering some invalid state. Can you replicate this behavior on your end?

The rest of the behavior appears to be correct, so once this quirk is localized I'll merge this PR along with the related mruby-zest-build one. Overall I think the new bank behavior is vastly more intuitive.

polluxsynth:
Thanks for testing!

When I've been testing this prior to submitting the PR, I've been triggering write failures in two ways: a) by attempting to write to a .xiz file which is read only (permissions 444) and b) by attempting to write to a directory which is read only (permissions 555). It looks as if Zest reports the write error consistently (i.e. it displays the "Failed to Save File" message in red on the second line in the console log in Zest); do you get that as well? I might have been unobservant about stray messages printed by Zyn though, but it sounds like in your case the resulting behavior as seen in Zest is incorrect - IIUC the error message does not get printed out in the UI? Plus that you get multiple messages per write attempt?

Do you have a log of the printouts from Zyn that I could have a look at to see if it matches what I get here?

polluxsynth:
I assume you mean the ten  Unknown address<BACKEND:online> '/alert:'  messages that are printed out by Zyn when starting Zest ? I honestly thought this was an old bug, or rather just an old cosmetic issue, as there are situations when Zyn emits similar Unknown address messages in response to other things done by Zest, but after testing both the master and development branches I see it's definitely brought on by the changes I've made so a clear regression. (Although oddly enough, it's the changes in Zest that does it - the messages come out whether or not Zyn is built from master or my development branch, while taking Zest back to master causes them to cease, no matter which Zyn version is built). So it would seem that my Zest changes bring out some unexpected behavior in Zyn ... at any rate I'll have a look at it.

it's definitely triggerby the changes in Zest (oddly enough I initially thought it was something in Zyn

fundamental:
Thinking about the bugs more in detail it's likely a consequence of the doReadOnlyOp() function rather than any modification and this time is just the most recent time that I've been watching the zyn output during that phase. Still a bug, but very likely not caused by any modification here.

> I honestly thought this was an old bug

Yep, that's an old bug caused by some incomplete details on the side of osc-bridge.

When a parameter is created on the zest side it's registered into the osc-bridge as something to watch for as well as something to cache. The cache says "I don't have a current value for this 'parameter'", then it goes about trying to fetch a value using the default approach, "just ask zyn for the value by sending a message which is the parameter's name with zero arguments" For any OSC path which is just an observable value (as opposed to a fetchable value), this creates an unknown address message with no args.

>  but is there some cases when the messages are bounced back to the sender?

Nope, just osc-bridge trying to fetch the value, which is how the vast majority of parameters function, but not 100% of them.

The more concerning cases of unknown addresses are when there are arguments attached to the message. For the included ones I was concerned about (which you can see on your end) zest requests a value, zyn generates a value, but the response isn't sent to the remote zest GUI and instead is sent back to the realtime thread, which isn't the expected behavior. The OSC handling graph shouldn't introduce any crashes, but it might result in a quirk or two.


Since the bugs seem to be a separate issue I'm merging in the improvements. Thanks for adding them :)

polluxsynth:
> When a parameter is created on the zest side it's registered into the osc-bridge as something to watch for as well as something to cache. The cache says "I don't have a current value for this 'parameter'", then it goes about trying to fetch a value using the default approach, "just ask zyn for the value by sending a message which is the parameter's name with zero arguments" For any OSC path which is just an observable value (as opposed to a fetchable value), this creates an unknown address message with no args.

Aha, so due to this behavior, it really is the `extern: "/alert"` which triggers the `Unknown address<BACKEND:online> '/alert'` messages then, as a result of osc-bridge attempting to fetch what would have been the corresponding parameter? 

Why are there 10 of them, is that an automatic retry timeout if no answer is received (the 100-ms-or-so between the messages would be consistent with this)?

> The more concerning cases of unknown addresses are when there are arguments attached to the message. For the included ones I was concerned about (which you can see on your end) zest requests a value, zyn generates a value, but the response isn't sent to the remote zest GUI and instead is sent back to the realtime thread, which isn't the expected behavior. The OSC handling graph shouldn't introduce any crashes, but it might result in a quirk or two.

I started debugging this a bit; I got as far as Master::applyOscEvent(): adding a printout just at the start of the routine, and after the call to ports.dispatch() further into the function, I get the following sequence just when attempting to save to a slot:

```
. . .
applyOscEvent: msg /vu-meter
applyOscEvent: after ports.dispatch: 1 matches
applyOscEvent: msg /active_keys
applyOscEvent: after ports.dispatch: 1 matches
applyOscEvent: msg /freeze_state
applyOscEvent: after ports.dispatch: 1 matches
applyOscEvent: msg /thaw_state
applyOscEvent: after ports.dispatch: 1 matches
applyOscEvent: msg /vu-meter
applyOscEvent: after ports.dispatch: 0 matches
Unknown address<BACKEND:online> '/vu-meter:ffffffffffffffffffffffffffffffffffffff'
applyOscEvent: msg /broadcast
applyOscEvent: after ports.dispatch: 0 matches
Unknown address<BACKEND:online> '/broadcast:'
applyOscEvent: msg /active_keys
applyOscEvent: after ports.dispatch: 0 matches
Unknown address<BACKEND:online> '/active_keys:FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF'
applyOscEvent: msg /vu-meter
applyOscEvent: after ports.dispatch: 1 matches
applyOscEvent: msg /active_keys
applyOscEvent: after ports.dispatch: 1 matches
. . .
```

So it would seem that ports.dispatch() isn't gobbling up the messages as it should for a short while, which I suspect is due to the preceding /freeze_state and /thaw_state which I'm supposing surrounds the actual save operation, although instinctively it would seem that the reaction is delayed, perhaps because of some queueing of messages between the various internal parts of zyn?

> Since the bugs seem to be a separate issue I'm merging in the improvements. Thanks for adding them :)

Thanks. I was starting to get ready for some debug sessions while at the same time getting to know more about the OSC message passing inside zyn... Some of the bugs seem to be newly triggered though which I still would feel is a bit worrying, as even if the bugs have been there before, they at least haven't been triggered before. Still, a couple of missing /vu-meter and /active_keys messages are in themselves no major cause for concern I assume.

fundamental:
> ... as a result of osc-bridge attempting to fetch what would have been the corresponding parameter?

Yep

> is that an automatic retry timeout if no answer is received (the 100-ms-or-so between the messages would be consistent with this)?

Sounds right

> Still, a couple of missing /vu-meter and /active_keys messages are in themselves no major cause for concern I assume.

Totally correct. They're fairly harmless, but indicate that there's a low chance that a more useful message could get skipped over.

> So it would seem that ports.dispatch() isn't gobbling up the messages as it should for a short while, which I suspect is due to the preceding /freeze_state and /thaw_state which I'm supposing surrounds the actual save operation, although instinctively it would seem that the reaction is delayed, perhaps because of some queueing of messages between the various internal parts of zyn?

It's going to be related to that process in some way or another. Messages shouldn't be lost when the system is put into a read-only state and there's a few queues to attempt to keep things sane, but I guess there must be at least one edge case missing.

polluxsynth:
Ok, so I found the issue, but I thought I'd bring it up here as the fix might not be as trivial as it should be.

In MiddleWare.cpp, in doReadOnlyOp, messasges from the backend are saved while waiting for /frozen_state from the backend. After the read only operation has been performed, the backend is thawed and the saved messages are sent on using code like:
```
    //Now to resume normal operations
    uToB->write("/thaw_state","");
    for(auto x:fico) {
        uToB->raw_write(x);
        delete [] x;
    }
```
The problem here is that all the saved messages are sent back to the backend using **uToB**->raw_write, when in fact they should be going in the other direction, using **bToU**->rawWrite. However, this means that middleware is impersonating the backend when calling bToU->raw_write which in the general case I'm assuming could lead to race conditions within that method. I'm thinking though that doing the insertion into the buffer _before_ sending /thaw_state should be safe enough, as the backend should not be sending messages while frozen.

fundamental:
I might have to reread all of my old code again to confirm this, but I think the original intent was to grab any mutations that could have been deferred and replay them for the backend.

So consider the following message sequence:

1. set volume to 25
2. save state (start a freeze)
3. set volume to 30
4. get key state
5. (end a freeze)

As a result the buffer going back to the middleware worker thread would look like:

1. volume has been set to 25
2. a frozen state has begun
3. message deferred since we're frozen "set volume to 30"
4. message deferred since we're frozen "get key state"
5. a frozen state has ended

Thus to get the backend at the state that would be implied from the original message sequence making the freezing hidden to users, you'd replay:

3. set volume to 30
4. get key state

which then generate:

- volume was set to 30
- current key state is XXXXXX

Not sure how well that matches the current state of the codebase though.

polluxsynth:
> I might have to reread all of my old code again to confirm this, but I think the original intent was to grab any mutations that could have been deferred and replay them for the backend.
> 
> So consider the following message sequence:
> 
>     1. set volume to 25
>     2. save state (start a freeze)
>     3. set volume to 30
>     4. get key state
>     5. (end a freeze)

Would all these messages come from the middleware (even if they did not originate there; things like volume changes would originate with the UI)? (It certainly looks like incoming MIDI CC messages are quietly discarded while frozen considering the first lines of the Master::setController() methods).

I'm probably too ignorant of the workings of the middleware, are there multiple threads which can send messages to the backend? Because otherwise it would seem that once in doReadOnlyOp(), there wouldn't be more messages sent to the backend after /state_freeze until the freeze ends.

> As a result the buffer going back to the middleware worker thread would look like:
> 
>     1. volume has been set to 25
>     2. a frozen state has begun
>     3. message deferred since we're frozen "set volume to 30"
>     4. message deferred since we're frozen "get key state"

Ok, so this is the point I seem to have missed: Messages received by the backend while frozen should be simply bounced back to the middleware, so that it can resend them after the freeze?

>     5. a frozen state has ended
> 
> Thus to get the backend at the state that would be implied from the original message sequence making the freezing hidden to users, you'd replay:
> 
>     3. set volume to 30
>     4. get key state
> 
> which then generate:
> 
>     * volume was set to 30
>     * current key state is XXXXXX

It would seem then that messages that are received from the backend prior to /state_frozen should really be handled as usual, either passed on directly or queued up and passed on later, and messages received after /state_frozen are deferred messages which should be queued up and sent back to the backend after the freeze?

Looking at doReadOnlyOp() it certainly looks like the middleware doesn't read anything from the backend after receiving /state_frozen. It only queues up messages that were received prior to that. My impression was that these were queued up so as not to cause further action towards the backend, as they might be if they were handled by bToUhandle(). 

I'm not sure i can get my head around all this, as I'm missing the original intent of the code here. I'll give it some more thought but perhaps this is something best left up to you for a rainy day.

fundamental:
> Would all these messages come from the middleware (even if they did not originate there; things like volume changes would originate with the UI)?

That would be a sequence sent from something external, like a GUI where messages generated within middleware would be within the parenthesis. It likely could use a full timing diagram to simplify discussions, but hopefully that clarification makes sense.

> I'm probably too ignorant of the workings of the middleware, are there multiple threads which can send messages to the backend?

Conceptually, there's:

a) The middleware thread
b) The realtime audio generation thread
c) optionally a thread for handling midi input if you've got mixed setups (e.g. ALSA in JACK out)

The only noteworthy variant is when padsynth is getting calculated as middleware spawns a pile of worker threads in that scenario.

> Ok, so this is the point I seem to have missed: Messages received by the backend while frozen should be simply bounced back to the middleware, so that it can resend them after the freeze?

Any messages received during a read only operation either need to be captured by middleware or if they're on the realtime side either dropped or sent right back to the middleware layer. I'd need to reread the exact design implemented to recall which of those two approaches was utilized.

> It would seem then that messages that are received from the backend prior to /state_frozen should really be handled as usual, either passed on directly or queued up and passed on later, and messages received after /state_frozen are deferred messages which should be queued up and sent back to the backend after the freeze?

That sounds correct.

> I'm not sure i can get my head around all this, as I'm missing the original intent of the code here. I'll give it some more thought but perhaps this is something best left up to you for a rainy day. :-)

Worst case while reviewing the PR you've opened I stumble across a different fix.

polluxsynth:
> > Would all these messages come from the middleware (even if they did not originate there; things like volume changes would originate with the UI)?
> 
> That would be a sequence sent from something external, like a GUI where messages generated within middleware would be within the parenthesis. It likely could use a full timing diagram to simplify discussions, but hopefully that clarification makes sense.

My understanding was that all OSC messages go via middleware to get to the backend (which I was equating with the RT audio generation thread but I'm getting the feeling the two are disjunct (but closely related)) but that is not the case then?

> > Ok, so this is the point I seem to have missed: Messages received by the backend while frozen should be simply bounced back to the middleware, so that it can resend them after the freeze?
> 
> Any messages received during a read only operation either need to be captured by middleware or if they're on the realtime side either dropped or sent right back to the middleware layer. I'd need to reread the exact design implemented to recall which of those two approaches was utilized.

Ok, I should check what actually happens when the read_only_fn() is called in doReadOnlyOp(). I was assuming that whatever messages were sent to middleware would simply remain in the queue is it wouldn't be reading it once the initial scan for /state_frozen was completed.

> > It would seem then that messages that are received from the backend prior to /state_frozen should really be handled as usual, either passed on directly or queued up and passed on later, and messages received after /state_frozen are deferred messages which should be queued up and sent back to the backend after the freeze?
> 
> That sounds correct.

Hm, this is even more confusing ... :-) It looks like doReadOnlyOp only queues up messages prior to /state_frozen; once it's seen that message is stops reading messages from the backend. What it does do is send the messages received prior to /state_frozen back to the backend (which is what the PR addresses).

> Worst case while reviewing the PR you've opened I stumble across a different fix.

Ok, good. The PR only addresses the case of messages received before /state_frozen, and from your description it would seem that it's correct to pass those on as usual, which is what the PR does. I can't see any code that deals with messages received after /state_frozen, unless that happens somewhere implicitly in the call to read_only_fn().

So at this stage, take the PR with a grain of salt, it only fixes the actual case I've seen where the /vu-meter and /active-keys messages from the backend on their way to the UI had been captured by doReadOnlyOp.

fundamental:
> My understanding was that all OSC messages go via middleware to get to the backend (which I was equating with the RT audio generation thread but I'm getting the feeling the two are disjunct (but closely related)) but that is not the case then?

External source (msg set 1) -> middleware receiving message -> middleware handling non-realtime portions of data (msg set 2) -> realtime receiving messages -> realtime responding to messages (msg set 3) -> middleware receiving messages -> middleware checking responses (msg set 4) -> middleware sending to remote -> remote receiving messages

and (msg set 2) -> responses from non-realtime data objects going to remote (msg set 5)

I think that covers all the noteworthy pathways.


> So at this stage, take the PR with a grain of salt, it only fixes the actual case I've seen where the /vu-meter and /active-keys messages from the backend on their way to the UI had been captured by doReadOnlyOp.

Will do.

PR #183, opened 2022-03-05
Update mod amount when depth/exp changed (#345)

polluxsynth:
When modwheel.depth or modwheel.exponential is changed, update the
modulation amount directly, rather than only when a new modwheel
value is received. This results in a more positive user experience,
otherwise it seems like changing the depth or exponential
parameters does not have any immediate effect.


fundamental:
Looks like a reasonable change to me. Merging enhancement :+1:

PR #184, opened 2022-03-05
Add modwheel and assocated parameters documentation

polluxsynth:
While trying to dig out how the modulation depth and exponential controls in the part panel actually work, I decided to try and document them for future reference

Please do read through to verify that I've gotten it right.


fundamental:
Consistent with what I'd expect and reading through src/Synth/ADnote.cpp and src/Params/Controller.cpp it looks like all descriptions are right.

PR #185, opened 2022-03-08
Update rest of controllers

polluxsynth:
Here I've done the same thing with the rest of the controller parameters as for the mod wheel. I've created two separate commits, one for controllers that have a depth parameter, and one for controllers that have enable toggles.

### DIFF
--- a/src/Params/Controller.cpp
+++ b/src/Params/Controller.cpp
@@ -65,15 +65,23 @@ const rtosc::Ports Controller::ports = {
     rParamI(pitchwheel.bendrange_down, rDefault(0),
         "Lower Range of MIDI Pitch Wheel"),
 #undef rChangeCb
-#define rChangeCb rChangeCbBase
+#define rChangeCb obj->setexpression(); rChangeCbBase
     rToggle(expression.receive, rShort("exp.rcv"), rDefault(true),
         "Expression MIDI Receive"),
+#undef rChangeCb
+#define rChangeCb obj->setfmamp(); rChangeCbBase
     rToggle(fmamp.receive,      rShort("fma.rcv"), rDefault(true),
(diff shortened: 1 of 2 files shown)


fundamental:
Looks good to me :+1:

polluxsynth:
Thanks!

polluxsynth:
I realized after some further testing that the behavior of this PR is not always what I intended, because for some of the CCs (read: volume and sustain), Part::SetController() checks the corresponding ctl.foo.receive variable, which I'd missed, before actually updating the synth engine. The end result is not half bad though: For sustain, any notes held by the sustain pedal when the receive switch is turned off actually continue to sustain (whereas newly played notes don't, as expected), and when the next sustain pedal CC is received, having a value of 0, it causes SetController() to release all keys that are sustained (if for some reason, a value of 127 would be sent again, the sustain will also be released, which technically is a bug, but unlikely to happen under normal circumstances). This sortof makes sense, because it means that notes that were sustained prior to the sustain receive being turned off will continue to be sustained until the sustain pedal is released.

For volume, it means that the volume remains at the last value set by the volume CC before volume receive was disabled, which might be preferable compared to the intended behavior, which was to make the volume jump up to max.

So quite frankly, I think SetController() should be left as it is as the resulting behavior is actually quite intuitive, but I just thought I'd mention it since I didn't originally realize that these CCs were handled differently, and just thought I'd make the behavior of all depth and receive enable parameters consistent with the PR.

fundamental:
I'm fine leaving the behavior as it stands. If a user is messing around with controller values while toggling on and off zyn's ability to receive those controllers I'm a-ok with saying 'that's undefined behavior'.

polluxsynth:
Ok good, thanks.

PR #186, opened 2022-03-13
Remove unused variable in demo version silencing code

nikolas:
This `tmp` var doesn't appear to be doing anything, and is throwing a compiler warning. Can it be removed?


fundamental:
Looks good to me and I like patches which result in less code :)

PR #188, opened 2022-03-13
globals.h: Increase smooth_float speed (#347)

polluxsynth:
Decrease the rate of smooth_float changes to make filter parameters
appear less sluggish. Otherwise it can take several seconds for
a parameter to reach the value set by a knob.

I don't know if this change has other consequences than the rate of parameter change when changing parameters in the UI, so this is more of a suggestion than a die hard fix.

### DIFF
--- a/src/globals.h
+++ b/src/globals.h
@@ -368,7 +368,7 @@ public:
         next_value = curr_value = value;
     };
     operator float() {
-        const float delta = (next_value - curr_value) / 128.0f;
+        const float delta = (next_value - curr_value) / 32.0f;
         curr_value += delta;
         return (curr_value);
     };


hselasky:
Looks good! I think I set 128.0f a long time ago as a good feeling value to avoid "square" waves appearing in the audio output when changing the parameters.

BTW: What buffer size are you using with ZynAddSubFX? Does it affect the sluggishness of these parameters?

polluxsynth:
I'm using a buffer size of 256 which I think is the default; as I recall Zynaddsubfx has issues with other buffer sizes.

Yes I think the resulting speed is related to the buffer size, however, as you said the main purpose is to avoid to much clickiness in the audio if a parameter is changed quickly. Say that the buffer size were smaller, then there would be more parameter updates per second to start with, and so the speed could be faster as it would have the same speed relative to the number of buffers processed per second.

fundamental:
> I'm using a buffer size of 256 which I think is the default; as I recall Zynaddsubfx has issues with other buffer sizes.

The only limit should be requiring that buffersize is divisible by 8 based on some optimizations I put in a number of years back.

> I'm using a buffer size of 256

I'll try to sanity check the change with respect to the plugin version buffersize which is a smaller (i.e. more expensive) buffer size of 32. If it sounds good there as well I'll merge and if not then perhaps there's a need for a buffersize scaling term.

polluxsynth:
Theorizing a bit, it would seem that with smaller buffer sizes, it stands to reason that a faster smooth_float speed would sound just as ok, as the amount of smoothing relative to the buffer rate would be the same (but I suppose that it might be perceived differently due to the different speed regardless of the theory of the underlying mechanism), on the other hand, it might be an advantage to have the same apparent rate regardless of the buffer size to make the synth behave consistently in all environments.

fundamental:
Messing around with Add Synth and some resonate low pass filters and sawtooth wave inputs. It seems like the change is good to my ears and baudline's plots. Tested with buffer size=32 and 1024. Merging :+1:

PR #189, opened 2022-03-15
Fix no analog filter q update

polluxsynth:
This fixes a problem with the Q for the Analog Filter not being updated unless the filter frequency is updated simultaneously, be it manually via the UI or MIDI CC, or via a sweeping envelope. The first commit fixes the problem in a rudimentary way, whereas the second one takes its cue from the setfreq method and attempts to only update the Q if it has changed significantly. I've chosen 10% as the limit, i.e. changes smaller than that will not be perpetrated. The final commit takes advantage of those situations when the coefficients are recalculated anyway (because the filter frequency is changed), and the latest set Q value is then taken regardless of if it's outside the 10% limit or not.

### DIFF
--- a/src/DSP/AnalogFilter.cpp
+++ b/src/DSP/AnalogFilter.cpp
@@ -35,6 +35,7 @@ AnalogFilter::AnalogFilter(unsigned char Ftype,
       stages(Fstages),
       freq(Ffreq),
       q(Fq),
+      newq(Fq),
      gain(1.0),
      recompute(true),
      freqbufsize(bufsize/8)
@@ -292,6 +293,8 @@ void AnalogFilter::setfreq(float frequency)
         freq = frequency;
         recompute = true;
     }
+    if (recompute)
+        q = newq;
     
     if (beforeFirstTick) {
         freq_smoothing.reset( freq );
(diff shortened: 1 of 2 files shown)


hselasky:
Hi,

There is also a type called:
src/globals.h:class smooth_float

Which serves a similar purpose, updating towards a given target value slowly.

--HPS

polluxsynth:
Yes, although the purpose for smooth_float and commit f8c4104 is different. f8c4104 is there to minimize the number coefficient calculations done, especially when a value changes slowly; a similar tradeoff is done by setfreq when the frequency changes less than 1 Hz. When adjusting parameters in Zyn, both frequency and Q are already handled as smooth_floats; the code here is just to avoid wasting too much CPU on coefficient calculations which will make barely audible differences.

You do raise an issue though, that when setting the parameter via MIDI CC#71 there is no smoothing done, for better or for worse. If CC#71 is updated manually, smoothing would be useful, but it might also be the result of automation when one wants to abruptly change the Q value at a specific point in a sequence. But that would have to be the subject of a different PR.

fundamental:
I'm not sure that the 10% tolerance is needed, but the original bug makes sense given the current state of code. It looks like the regression was introduced by myself when switching code over to ModFilter which likely stripped out old setq() calls in favor of the setfreq_and_q() calls.

Thanks for the resolution :+1:

fundamental:
In the millisecond before hitting rebase and merge I thought about SVFilter. Looks like the same issue exists there as well as FormantFilter, but not Moogfilter. Do you want to have those in this PR as well? It would more or less be the same logic.

polluxsynth:
I'll have a look at it. When I tested it now though, SVF seems fine, but for the formant filter, neither the Q UI knob nor external CC seems to have any effect after the envelope has stopped, so it's slightly different from the analog filter behavior, where the Q knob was ok but not external MIDI CC.

fundamental:
Ah ok. Thanks for the check. That's what I get for a partial code reading without testing the edge case myself.

Merging :+1:

PR #190, opened 2022-03-20
SVFilter.cpp: Avoid clicks when UI updates freq, Q or gain (#348)

polluxsynth:
In the setstages() method, only call cleanup() when needed,
i.e. when the number of stages is actually changed (same as
in AnalogFilter.cpp).

(The reason this is a problem is because setstages() is
called whenever the frequency, Q, filter tracking or gain
is changed in the UI).

### DIFF
--- a/src/DSP/SVFilter.cpp
+++ b/src/DSP/SVFilter.cpp
@@ -151,9 +151,11 @@ void SVFilter::setstages(int stages_)
 {
     if(stages_ >= MAX_FILTER_STAGES)
         stages_ = MAX_FILTER_STAGES - 1;
-    stages = stages_;
-    cleanup();
-    computefiltercoefs();
+    if (stages != stages_) {
+        stages = stages_;
+        cleanup();
+        computefiltercoefs();
+    }
 }
 
 float *SVFilter::getfilteroutfortype(SVFilter::fstage &x) {


fundamental:
I can confirm this fixes the bug. Guess I overlooked that call which is likely coming from the ModFilter.cpp level when the timestamp on the filterparameters get updated.

Merged :+1:

polluxsynth:
On a deeper note (sorry about the pun), how come there is this logic to send updates to several parameters when only one is updated? Without studying the code in detail, it would seem to be a high level attempt at getting an update to one filter parameter to get the code to recalculate the coefficients based on all parameters, but that is also done by the actual filter implementations (and in some cases, like receiving CC changes, that is the only place that it's done); shouldn't it be done in one or the other place but not both?

fundamental:
> how come there is this logic to send updates to several parameters when only one is updated?

I'm assuming you're referring to the timestamp based updates of some of the src/Synth/ code pulling from src/Params/ code. In these instances it was generally done to simplify the update logic. There's not much in the way of a performance cost to redo some calculations when a semi-related parameter changes. Going a step further I'd expect `if(param2 != param2old && all_other_parameters_const) specialized_update()` to be slower given all of the additional branching.

> t would seem to be a high level attempt at getting an update to one filter parameter to get the code to recalculate the coefficients based on all parameters, but that is also done by the actual filter implementations 

Now I'm confused. Are you talking about the Modfilter code and the individually smoothed floats within the filters? I wouldn't be opposed to simplifying some of the code, though I'm not sure it would make too much of a difference in sound/performance.

> shouldn't it be done in one or the other place but not both?

Assuming I'm interpreting the question right, it should be done in one place if one place is sufficient.

polluxsynth:
> > how come there is this logic to send updates to several parameters when only one is updated?
> 
> I'm assuming you're referring to the timestamp based updates of some of the src/Synth/ code pulling from src/Params/ code. In these instances it was generally done to simplify the update logic. There's not much in the way of a performance cost to redo some calculations when a semi-related parameter changes. Going a step further I'd expect `if(param2 != param2old && all_other_parameters_const) specialized_update()` to be slower given all of the additional branching.
> 
> > t would seem to be a high level attempt at getting an update to one filter parameter to get the code to recalculate the coefficients based on all parameters, but that is also done by the actual filter implementations
> 
> Now I'm confused. Are you talking about the Modfilter code and the individually smoothed floats within the filters? I wouldn't be opposed to simplifying some of the code, though I'm not sure it would make too much of a difference in sound/performance.

Your confusion is probably because I only half know what I'm talking about...and I may have misunderstood something vital...

Taking he case in point as an example, we have a change to the filter frequency come in, usually from the UI, as an OSC command. Now, the upper level makes a decision to call update functions in Modfilter not just to update the frequency but also the other parameters, i.e. the upper level calls both setfreq() and setstages() among others. The individual Modfilter implementations (AnalogFilter etc) first get a call to setfreq() then immediately after a call to setstages(). It would seem to me that a 'one OSC parameter in, one Modfilter parameter change out' paradigm would have sufficed. There is already a table of OSC ports and what to do when something happens on each of them, so I was expecting that a change in one parameter, say a filter freq change, would eventually work its way to setfreq() and nothing else, rather than first collecting in a 'here we gather all filter parameter updates' thing which then calls a set of filter implementation update methods such as setfreq(). But perhaps this would become clearer if I'd actually taken to writing an implementation for something like this and experienced the actual issues at hand rather than just theorizing about them.

fundamental:
So, you've got Synth objects which contain the DSP, you've got parameter objects which specify the characteristics which each DSP class should be _initialized_ with. More recently (past few years) the additional behavior is that DSP objects get to have references to the parameter objects and are able to see when the parameter objects have been modified. Without keeping copies of the old parameter values in the referenced parameter object, you're stuck with generally updating a full block of information and on the DSP side skipping steps if they're seen to be no-ops.

Basically just another quirk of how zyn's design has been upgraded to support live parameter updates where in the past it only supported reflecting changes when a new note was pressed i.e. when new src/Synth/* objects were created and initialized with the parameter objects which the users were editing.

polluxsynth:
So, in practice a result of Zyn growing organically, simply put.

PR #191, opened 2022-03-23
Part, NotePool: Use correct method to release keys (#346)

polluxsynth:
Use the correct NotePool method to release keys, which set the
released keys in the RELEASED state, rather than bypassing the
note pool and directly calling the releasekey() method for the
engine(s).

### DIFF
--- a/src/Containers/NotePool.cpp
+++ b/src/Containers/NotePool.cpp
@@ -465,6 +465,17 @@ void NotePool::releasePlayingNotes(void)
     }
 }
 
+void NotePool::releaseSustainingNotes(void)
+{
+    for(auto &d:activeDesc()) {
+        if(d.sustained()) {
+            d.setStatus(KEY_RELEASED);
+            for(auto s:activeNotes(d))
+                s.note->releasekey();
+        }
+    }
+}
+
 void NotePool::release(NoteDescriptor &d)
 {
     d.setStatus(KEY_RELEASED);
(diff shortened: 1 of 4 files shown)


fundamental:
There's a minor difference in terms of the status bit checks for ReleaseAllKeys(), but from what I can tell it's not an edge case that impacts anything of note and the new solution does utilize the notepool more cleanly.

Merged :+1:

polluxsynth:
Thanks!

It seemed to me that the status bit checks in releasePlayingNotes() actually were more appropriate, since they do release keys that are actually playing, whereas the old version would have set keys that were entombed to released, although obviously (since it doesn't seem to cause problems) that was not a problem in itself.

It struck me by the way both when when looking at the code and after playing Zyn with someone elses Roland keyboard yesterday that there is a bug in the handling of the all notes off message: In Zyn it unconditionally releases the keys, which will cause problems with Roland keyboards, as they emit an 'all notes off' message whenever all keys are released on the keyboard, regardless of the sustain pedal status. In essence 'all notes off' is a bit of a misnomer, it doesn't mean 'turn all voices off now', rather, 'there are no more keys pressed on the keyboard, so sending this just to be sure in case you missed one or more all note offs'.

I don't think I'll raise an issue for this, just send a new PR, if that's ok. The bug is blatently obvious is you use Zyn with a Roland keyboard and sustain pedal (actually, I don't know if they still have this behavior; the older ones like Alpha Juno 2 and D-50 definitely do. It's been the subject of several Roland OS ROM hacks on my part to get rid of the behavior, because it wrecks havoc with some sequencers, so I didn't notice it at home with my modified Juno 2).

fundamental:
> What corner case were you thinking of?

Nothing harmful. I was just noting that in the code review there was behavior change rather than just refactoring. I'd agree with you that the new behavior appears to be more correct.

PR #192, opened 2022-03-24
Part.cpp: Fix all notes off

polluxsynth:
When the MIDI All Notes Off is received, this is equated to a
note off for all currently playing keys, which is not the
same as releasing all playing voices, as the latter does not
take into consideration the sustain pedal state.

This is noticable when playing Roland keyboards, which habitually
emit an All Notes Off message whenever all keys are released.

Tested with a Roland D-10 synth, which has the typical Roland behavior.

### DIFF
--- a/src/Misc/Part.cpp
+++ b/src/Misc/Part.cpp
@@ -686,7 +686,7 @@ void Part::NoteOff(note_t note) //release the key
     for(auto &desc:notePool.activeDesc()) {
         if(desc.note != note || !desc.playing())
             continue;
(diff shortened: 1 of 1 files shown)


fundamental:
I wasn't entirely sure if this behavior was just a quirk of your specific keyboard(s) or not, so I asked in #lad. It seems like your interpretation of the all-notes-off message is consistent with some other hardware and software, therefore it's reasonable to call zyn's old behavior as buggy. Looks like the new implementation should behave as you had expected as described in  #191 

Merged :+1:

PR #195, opened 2022-04-11
MiddleWare: Handle queued up messages during doReadOnlyOp

polluxsynth:
Correctly handle messages that have been queued up from the backend
while performing a doReadOnlyOp, rather than sending them
back to backend.

This gets rid of messages like the following when doing a write
operation from the UI:

    Unknown address<BACKEND:online> /vu-meter:ffffffffffffffffffffffffffffffffffffff
    Unknown address<BACKEND:online> /broadcast:
    Unknown address<BACKEND:online> /active_keys:FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF

### DIFF
--- a/src/Misc/MiddleWare.cpp
+++ b/src/Misc/MiddleWare.cpp
@@ -2073,7 +2073,7 @@ void MiddleWareImpl::doReadOnlyOp(std::function<void()> read_only_fn)
     //Now to resume normal operations
     uToB->write("/thaw_state","");
     for(auto x:fico) {
-        uToB->raw_write(x);
+        bToUhandle(x);
         delete [] x;
     }
 }
@@ -2180,7 +2180,7 @@ bool MiddleWareImpl::doReadOnlyOpNormal(std::function<void()> read_only_fn, bool
         //Now to resume normal operations
         uToB->write("/thaw_state","");
         for(auto x:fico) {
-            uToB->raw_write(x);
+            bToUhandle(x);
             delete [] x;
         }
         return false;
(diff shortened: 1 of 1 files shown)


fundamental:
Double checking the source, there's nothing fancy on the end of the realtime thread while the state variable of frozenState is true. On the side of Middleware any messages heading to the realtime side are blocked by virtue of it being a single thread design which polls for new network events and doReadOnlyOp blocks until the operation is complete.

Thus it does look like you're correct in that any events appearing in bToU are either status information or messages which should be handled correctly. As such they should be handled the same as in tick() whenever the read only operation is complete.

So, yep looks like you figured out the right solution unless I've missed an old implementation decision of mine.

Thanks for tracking down the issue :+1:

polluxsynth:
Great, thanks!

As noted in https://github.com/zynaddsubfx/zynaddsubfx/pull/181#issuecomment-1100759787, I've been considering and to some extent evaluating an alternative fix, given that the one in this PR is the correct one, namely to add a "read lookahead" method to ThreadLink, whereby it would be possible to read ahead without actually advancing the normal read pointer. That saves actually stacking up the incoming messages on a separately allocated queue - basically, doReadOnlyOp() reads ahead, looking for /state_frozen, but leaving the messages in the buffer so that tick() can pick them up later.

It turns out it doesn't require a lot of code changes to ThreadLink, and seems to me to be a cleaner fix on the whole, although of course more intrusive when it comes to rtosc. I would have to add tests to rtosc to verify the correct behavior.

One downside is that /state_frozen remains in the queue and is also picked up subsequently by tick(), although that doesn't seem to have any ill effects; if nothing else, I'm guessing one could add a dummy port so that the message would be handled in a well-defined manner.

PR #200, opened 2022-05-08
Readonlyops use lookahead

polluxsynth:
Use the newly introduced read lookahead feature in ThreadLink in doReadOnlyOp and friends.

### DIFF
--- a/src/Misc/MiddleWare.cpp
+++ b/src/Misc/MiddleWare.cpp
@@ -2047,20 +2047,15 @@ void MiddleWareImpl::doReadOnlyOp(std::function<void()> read_only_fn)
     assert(uToB);
     uToB->write("/freeze_state","");
 
-    std::list<const char *> fico;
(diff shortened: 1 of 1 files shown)


polluxsynth:
Looks like the CI process failed because it did not pull in the latest rtosc source. I'm guessing I've messed up with the commit that attempts to do that, or is there some separate setup that need to be done in order for that to happen?

fundamental:
Everything looks reasonable with the addition here and it does simplify the flow. There is the absolutely minuscule risk that the backend runs out of space on the return ringbuffer, but it's not realistic for me to be concerned about that IMO as it's only one message which gets written. As per the Arch Linux test, that's based around the assumption that rtosc is finalized and _should_ be devendored, which well is not the case (IMO), so I'll disable it until there's another joint zyn+rtosc release from which point it can be re-enabled.

PR #201, opened 2022-05-08
Add likely missing else

polluxsynth:
Slightly unsure of this one, but forward and broadcast seem to be mutually exclusive, and the layout of the code indicates that there is a missing else.

It doesn't seem to fix any observable bug though, so if there's any doubt just drop this PR. I just came across it while looking debugging and thought I'd mention it is it looks odd.


fundamental:
I'd say this change is right on the money. Merged :+1:

polluxsynth:
Thanks!

PR #205, opened 2022-07-18
Pfilter fc ctl bypass

friedolino78:
this commit makes it possible to disable the filter cutoff midi control (CC74) per voice. This is good for pitch tuned (comb) filters.

### DIFF
--- a/src/Synth/ADnote.h
+++ b/src/Synth/ADnote.h
@@ -187,7 +187,7 @@ class ADnote:public SynthNote
             /* Filter Bypass */
             int filterbypass;
             /* Filter Fq Control Bypass */
-            int filterFqCtlBypass;
+            int filterFcCtlBypass;
 
             /* Delay (ticks) */
             int DelayTicks;
--- a/src/Synth/ADnote.h
+++ b/src/Synth/ADnote.h
@@ -186,6 +186,8 @@ class ADnote:public SynthNote
 
             /* Filter Bypass */
             int filterbypass;
+            /* Filter Fq Control Bypass */
+            int filterFqCtlBypass;
 
             /* Delay (ticks) */
             int DelayTicks;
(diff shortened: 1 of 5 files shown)


fundamental:
I don't mind this feature only applying to addsynth voice filters, but I'm just checking the intention is to omit the new behavior on addnote/padnote/subnote global level filters, right?

I also spent a bit of time verifying that replacing relfreq with 0.0f is correct and based on the math in Controller.cpp, it corresponds with the knob being at a value of 64, which looks great to me.

Once the above question is answered I'm in favor of squashing and merging :+1:

friedolino78:
> I don't mind this feature only applying to addsynth voice filters, but I'm just checking the intention is to omit the new behavior on addnote/padnote/subnote global level filters, right?

actually i don't know what new behavior you refer to.
I just want to be able to select the filters where relfreq should be applied to. right now relfreq modifies all filters in the same way.
 
> I also spent a bit of time verifying that replacing relfreq with 0.0f is correct and based on the math in Controller.cpp, it corresponds with the knob being at a value of 64, which looks great to me.

I deduced the neutral value for relfreq from the modfilter: Fc_mod = Fc + relfreq + tracking;
great, that we come to the same result from different directions.

> Once the above question is answered I'm in favor of squashing and merging +1

src/Synth/ADnote.h
```diff
@@ -186,6 +186,8 @@ class ADnote:public SynthNote
 
             /* Filter Bypass */
             int filterbypass;
+            /* Filter Fq Control Bypass */
+            int filterFqCtlBypass;
```
friedolino78, line 190: reminder: refactor to filterFcCtlBypass

fundamental:
> For my use case it is sufficient to disable relfreq in voices. But now I think of making it more generic by adding the same port/feature to global filters, too.

Just checking, is this comment indicating you want to expand this PR or have this merged before working on any other filters?

friedolino78:
> > For my use case it is sufficient to disable relfreq in voices. But now I think of making it more generic by adding the same port/feature to global filters, too.
> 
> Just checking, is this comment indicating you want to expand this PR or have this merged before working on any other filters?

merge this first.

fundamental:
Merged :+1:

PR #208, opened 2022-08-18
fix lfo parameter limit

friedolino78:
make freq randomization reatlime

### DIFF
--- a/src/Synth/LFO.cpp
+++ b/src/Synth/LFO.cpp
@@ -279,7 +279,7 @@ float LFO::lfoout()
         phase += phaseInc;
     else {
         const float tmp = (incrnd * (1.0f - phase) + nextincrnd * phase);
-        phase += phaseInc * limit(tmp, 0.0f, 1.0f);
+        phase += phaseInc * tmp;
     }
     if(phase >= 1) {
         phase    = fmod(phase, 1.0f);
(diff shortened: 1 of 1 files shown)


hselasky:
Spelling error in commit message: "reatlime" ?

hselasky:
Can you explain why 4.0f is used instead 1.0f. I thought phases were always [0.0f ... 1.0f>

fundamental:
I still have to double check the origin of the limit statement, but IMO that's a separate issue from the realtime parameter responses.

@hselasky Some discussion outside of the issue tracker made it look like when the frequency randomization is moved to larger and larger values the limit statement was clipping the values making it less random not more random as would be intended. I haven't worked through the numbers myself to validate the exact domain of each of the variables, but that is what was discussed. I'd say phase should be limited 0.0..1.0, but that's a limitation on phaseInc*tmp, not on tmp itself.

@friedolino78 I have no objection to the second line in the commit, so that should be an easy merge, though I'd fix the typo that HPS mentioned.

friedolino78:
> Can you explain why 4.0f is used instead 1.0f. I thought phases were always [0.0f ... 1.0f>

limited is only a factor to phaseInc.

JohannesLorenz:
Based on the discussion here is my proposal:

1. The first changed line (`lfofreqrnd = powf(lfopars.Pfreqrand / 127.0f, 2.0f) * 4.0f;`) is acceptable. It can be in a separate commit, as it already is, or in a separate PR - both is fine and can be merged.
2. The second line (`phase += phaseInc * limit(tmp, 0.0f, 4.0f)`) needs to be well documented. Why do you change this after 8 years, and why `4.0`? Can you please document this in the code and/or in the commit message?

friedolino78:
> 2\. The second line (`phase += phaseInc * limit(tmp, 0.0f, 4.0f)`) needs to be well documented. Why do you change this after 8 years, and why `4.0`? Can you please document this in the code and/or in the commit message?

the value of tmp comes from this line:
nextincrnd = powf(0.5f, lfofreqrnd) + RND * (powf(2.0f, lfofreqrnd) - 1.0f);
with lfofreqrnd at its maximum value 4.0, the value range goes from:
0.5^4 + 0
0.0725

to:
0.5^4 + 1.0 * (2^4 - 1)
0.0725 + 15
15.0725

phase += phaseInc * limit(tmp, 0.0f, 1.0f);
saturates >90% of the time. i guess it's simply a bug and was meant to saturate phase

friedolino78:
the last commit fixes the problem, that the original random factor changed the mean frequency.

friedolino78:
now it's rebased to current master. And all test pass.

JohannesLorenz (approved):

PR #210, opened 2022-09-03
add tan waveshaper

friedolino78:
features a new waveshaper function tan for distortion and oscilgen


fundamental:
> // TBD: tuning the function to

tuning the function to what?

friedolino78:
> > // TBD: tuning the function to
> 
> tuning the function to what?

tuning to equal gain through the parameter range. right now it has a peak at 64.
maybe two linear segments could sove this.

fundamental:
It looks like the FLTK gui mods were only to the simple GUI. That had me confused for a few minutes.

fundamental:
Looks like the zest PR didn't have the Hysteresis distortion function.

IMO the hysteresis function is too aggressive (with respect to the drive parameter) and it appears to be uniquely poor with respect to the DC parameter in that it can easily result in a distortion which produces a large DC bias in the output (even for zero valued inputs).

So, I'd say the first distortion appears good to go with the missing FLTK GUI non-simple UI dropdown and the second one needs more work to tune the parameter range for drive and something to resolve the issues with respect to the DC bias.

friedolino78:
> IMO the hysteresis function is too aggressive (with respect to the drive parameter) and it appears to be uniquely poor with respect to the DC parameter in that it can easily result in a distortion which produces a large DC bias in the output (even for zero valued inputs).

indeed the dc output compensation for hyst is missing. Now I found out, that it is broken in arctan, too.

friedolino78:
things have been fixed. next round for reviewing.

fundamental:
DC offsets appear to be fixed. The tan feels like it can be a bit overly aggressive at higher drives so maybe it should be limited to +-1.3f instead of +-1.5f? I also explored changing the offset range only for tan because then it's possible to shift the signal more into the non-linear region, but I can't say it really impacts usability that much.

As per the discussed regression test, I've attached it, though since I altered the tan code locally I'd expect the test to fail specifically on those elements. Regenerating the coefficients and testing before/after this PR should be fairly straight forward.
[DistortionTest.txt](https://github.com/zynaddsubfx/zynaddsubfx/files/11857097/DistortionTest.txt)

friedolino78:
> The tan feels like it can be a bit overly aggressive at higher drives so maybe it should be limited to +-1.3f instead of +-1.5f?

Unfortunately that limiter is hard clipping the input to prevent clipping at the output. quite a paradox intervention.
But in lack of other ideas to tame tan it might be a better tradeoff.

fundamental:
As per discussions outside of the issue tracker it sounds like there's still normalization work to be done?

friedolino78:
i replaced the tan function by a dual cos function for having a similar feature at zerocrossing without the normalizing problems.

JohannesLorenz:
@friedolino78 can you please solve the merge conflicts?

fundamental:
Merge conflicts fixed via https://github.com/zynaddsubfx/zynaddsubfx/commit/d33b55c3a656741c089e3742e40d41845438cf86 . AFAIK this is good to go. Let me know if this looks sane and I'll shift it to the master branch.

PR #218, opened 2022-12-12
Fix distortion effect multiplier

JohannesLorenz:
This fixes a bug introduced in 85ed5029787cbbc3b4ea50bc18b9182526d923eb. That commit accidentally multiplied the preset value for volume by 3/2, instead of by dividing it ("lowering") by 3/2.

This fixes knobs displaying values like "190" and presets being too loud.


JohannesLorenz:
@hselasky It would be nice if you can look at this one-liner, since you know the code already :smiley:

hselasky:
The patch makes sense. You can also say, make code match description.

hselasky (approved):
Looks good!

JohannesLorenz:
@hselasky Thanks for the review.

PR #219, opened 2022-12-13
Fix sympathetic effect multiplier

JohannesLorenz:
Analogous implementation to
4e0b22f267f4254ad9deed74856d9155aef55848, but this time, for the Sympathetic effect.


friedolino78 (commented):
looks plausible

PR #220, opened 2022-12-21
Use logmin in Schema export and in some ports

JohannesLorenz:
This commit exports the "logmin" value in the Schema and uses the new `rLogWithLogmin` for some ports.

This also fixes default values which were originally below the `[min, max]` range. Before this commit, you could not set these values using the OSC ports, because of `rLimit`.

Corresponding UI PR: mruby-zest/mruby-zest-build#89.


fundamental:
Looks all good to me. Thanks for the enhancement :+1:

PR #221, opened 2023-02-01
Use DPF's CMake module to build plugins

AnClark:
DPF now has a CMake module for building musical plugins easily. **Simply using a function `dpf_add_plugin()`**, so we can save a bunch of energy in writing rules by hand, and it will be easier to maintain.

Original rules included building DPF main files, UI files and lv2-ttl-generator by hand, as well as linking against right libraries. But now `dpf_add_plugin()` can do these for us normatively, and of course, automatically.

Thanks to upgrades of DPF, this patch also bring ZynAddSubFX (as well as its FX plugins) more format supports. No more being limited to LV2 and VST2, **now ZynAddSubFX supports VST3 and CLAP**, the two modern plugin standards.

What's more, newer versions of DPF has better performance on DGL, so UI can run much more fluently. This is an example: <https://github.com/DISTRHO/imgui-template-plugin>.

**Features:**

- New plugin format support: VST3, CLAP
- Use `dpf_add_plugin()` to build plugins

**Notice:**

Only tested bare UI (NTK, FLTK).

Working with synthesizer side (`/usr/bin/zynaddsubfx`) and Zest has not yet tested.

**Screenshots:**

FLTK UI:

image

image

NTK UI:

image

image


fundamental:
FYI the port checker test has been known to raise false positives and no changes in this PR should be able to cause a regression that would impact that specific test.

AnClark:
This PR does not touch sources related to OSC.

I've run test locally, and all test passed:

```
Test project /home/anclark/Sources/zynaddsubfx/build
      Start  1: rtosc-time
 1/56 Test  #1: rtosc-time .......................   Passed    0.00 sec
      Start  2: osc-spec
 2/56 Test  #2: osc-spec .........................   Passed    0.00 sec
      Start  3: simple-messages
 3/56 Test  #3: simple-messages ..................   Passed    0.00 sec
      Start  4: null-messages
 4/56 Test  #4: null-messages ....................   Passed    0.00 sec
      Start  5: bundles
 5/56 Test  #5: bundles ..........................   Passed    0.00 sec
      Start  6: patterns
 6/56 Test  #6: patterns .........................   Passed    0.00 sec
      Start  7: nested-bundles
 7/56 Test  #7: nested-bundles ...................   Passed    0.00 sec
      Start  8: fat-message
 8/56 Test  #8: fat-message ......................   Passed    0.01 sec
      Start  9: empty-strings
 9/56 Test  #9: empty-strings ....................   Passed    0.00 sec
      Start 10: message-alignment
10/56 Test #10: message-alignment ................   Passed    0.00 sec
      Start 11: test-arg-iter
11/56 Test #11: test-arg-iter ....................   Passed    0.00 sec
      Start 12: liblo
12/56 Test #12: liblo ............................   Passed    0.37 sec
      Start 13: version
13/56 Test #13: version ..........................   Passed    0.00 sec
      Start 14: util
14/56 Test #14: util .............................   Passed    0.01 sec
      Start 15: arg-val-cmp
15/56 Test #15: arg-val-cmp ......................   Passed    0.01 sec
      Start 16: arg-val-math
16/56 Test #16: arg-val-math .....................   Passed    0.01 sec
      Start 17: thread-link-test
17/56 Test #17: thread-link-test .................   Passed    0.00 sec
      Start 18: test-midi-mapper
18/56 Test #18: test-midi-mapper .................   Passed    0.01 sec
      Start 19: metadata
19/56 Test #19: metadata .........................   Passed    0.01 sec
      Start 20: apropos
20/56 Test #20: apropos ..........................   Passed    0.01 sec
      Start 21: pretty-format
21/56 Test #21: pretty-format ....................   Passed    0.01 sec
      Start 22: default-value
22/56 Test #22: default-value ....................   Passed    0.01 sec
      Start 23: headerlib
23/56 Test #23: headerlib ........................   Passed    0.05 sec
      Start 24: test-walker
24/56 Test #24: test-walker ......................   Passed    0.01 sec
      Start 25: walk-ports
25/56 Test #25: walk-ports .......................   Passed    0.01 sec
      Start 26: path-search
26/56 Test #26: path-search ......................   Passed    0.01 sec
      Start 27: undo-test
27/56 Test #27: undo-test ........................   Passed    0.01 sec
      Start 28: sugar
28/56 Test #28: sugar ............................   Passed    0.01 sec
      Start 29: typed-template-test
29/56 Test #29: typed-template-test ..............   Passed    0.01 sec
      Start 30: path-collapse
30/56 Test #30: path-collapse ....................   Passed    0.01 sec
      Start 31: test-automation
31/56 Test #31: test-automation ..................   Passed    0.01 sec
      Start 32: test-port-checker
32/56 Test #32: test-port-checker ................   Passed    0.56 sec
      Start 33: AdNoteTest
33/56 Test #33: AdNoteTest .......................   Passed    0.36 sec
      Start 34: AllocatorTest
34/56 Test #34: AllocatorTest ....................   Passed    0.06 sec
      Start 35: ControllerTest
35/56 Test #35: ControllerTest ...................   Passed    0.01 sec
      Start 36: EchoTest
36/56 Test #36: EchoTest .........................   Passed    0.02 sec
      Start 37: EffectTest
37/56 Test #37: EffectTest .......................   Passed    0.05 sec
      Start 38: KitTest
38/56 Test #38: KitTest ..........................   Passed    0.24 sec
      Start 39: MemoryStressTest
39/56 Test #39: MemoryStressTest .................   Passed    0.39 sec
      Start 40: MicrotonalTest
40/56 Test #40: MicrotonalTest ...................   Passed    0.01 sec
      Start 41: MsgParseTest
41/56 Test #41: MsgParseTest .....................   Passed    0.01 sec
      Start 42: OscilGenTest
42/56 Test #42: OscilGenTest .....................   Passed    0.71 sec
      Start 43: PadNoteTest
43/56 Test #43: PadNoteTest ......................   Passed    0.65 sec
      Start 44: PortamentoTest
44/56 Test #44: PortamentoTest ...................   Passed    0.02 sec
      Start 45: RandTest
45/56 Test #45: RandTest .........................   Passed    0.01 sec
      Start 46: SubNoteTest
46/56 Test #46: SubNoteTest ......................   Passed    0.05 sec
      Start 47: TriggerTest
47/56 Test #47: TriggerTest ......................   Passed    0.03 sec
      Start 48: UnisonTest
48/56 Test #48: UnisonTest .......................   Passed    0.14 sec
      Start 49: WatchTest
49/56 Test #49: WatchTest ........................   Passed    0.02 sec
      Start 50: XMLwrapperTest
50/56 Test #50: XMLwrapperTest ...................   Passed    0.01 sec
      Start 51: PluginTest
51/56 Test #51: PluginTest .......................   Passed    7.33 sec
      Start 52: MiddlewareTest
52/56 Test #52: MiddlewareTest ...................   Passed    2.16 sec
      Start 53: MqTest
53/56 Test #53: MqTest ...........................   Passed    0.02 sec
      Start 54: MessageTest
54/56 Test #54: MessageTest ......................   Passed    1.48 sec
      Start 55: PortChecker
55/56 Test #55: PortChecker ......................   Passed    9.58 sec
      Start 56: SaveOsc
56/56 Test #56: SaveOsc ..........................   Passed    3.17 sec

100% tests passed, 0 tests failed out of 56

Total Test time (real) =  27.74 sec
```

AnClark:
What's more, plugin output directory has changed to `${CMAKE_BINARY_DIR}/bin`. So, if merged, Zyn-Fusion build scripts will need to modify as well.

AnClark:
I wonder if it is OK to go forward. At least NTK and FLTK versions work well.

fundamental:
I'm planning on building the code for some quick sanity checks this evening, but based upon my previous reading of the diffs this should likely be a quick merge. Are you interested in making the modifications on the build repo as well or will that get assigned to someone else?

fundamental:
I can confirm on my system the lv2-ttl-generator does not build due to the removal of the -ldl linker flag, that should be an easy fix. Do you know if the updated DPF should handle the OSX code which was removed within this PR?

AnClark:
> Are you interested in making the modifications on the build repo as well or will that get assigned to someone else?

Do you mean `zyn-fusion-build`? Modifying those makefiles should be easy. I can handle this.

I've ported several plugins to DPF. `lv2-ttl-generator` should work in both Windows ans Linux. But I don't have a macOS computer (I can't afford them now :cry: )  You can check `cmake/DPF-plugin.cmake` in DPF source tree (in `develop` branch).

AnClark:
By the way, I have a CMake-based Zyn-Fusion build script, which manages dependencies better: <https://github.com/anclark/zyn-fusion-build/tree/new-generation>. Would you love to take a look?

fundamental:
To confirm, this patch set should be complete from your perspective, right?

AnClark:
> To confirm, this patch set should be complete from your perspective, right?

During tests before, I've found several bugs when testing cross-compiling on my Arch Linux, and had made a bundle of Force-pushes to fix them.

Let me have another check.

AnClark:
OK, now things works well on ZynAddSubFX compilation (Zyn only):

- **On Linux (native build):** Build passed, and audio can play in all formats and all UI types.
- **On Windows (cross-build):** Build passed via zyn-fusion-build (`Makefile.windows.mk`).
- **On Msys2:** Not tested yet.

Shall I test it on native Windows? Or would you love to assign someone else to test?

fundamental:
Having a test of the binaries on windows would be great. I don't locally have a pure windows machine, so I mostly have to use wine and then trust that things haven't broken massively between updates. Based on the discussion about the FLTK UI it sounds like there might be some delays in merging this patchset. It's a positive though because it sounds like this will help upstream DPF become more feature complete :)

AnClark:
Great! Both Zyn and DPF could benefit from here. I'll edit this PR tonight (BJT) according to falkTX's clarifications.

BTW, in my ThinkPad X201T (with Manjaro), ZynAddSubFX with bare Zest support (Zest is not loaded actually) is slow. How about yours?

AnClark:
Hi, Mark! Is it possible to go forward?

fundamental:
Thanks for the reminder. I swear last time I checked this PR the EXTERNAL_UI flag was still flipped the other way, but it looks like that was corrected (maybe it was github's UI interacting with the force pushes in a strange way). It looks like things are good, but since I'm half-awake right now I'll schedule some time tomorrow morning to give the code a last sanity check before merging.

AnClark:
> I'll schedule some time tomorrow morning to give the code a last sanity check before merging.

Great!

A further sanity check is necessary, since I've partially checked NTK and FLTK part (I was busy with my exams this month), but not the full Zest. I will test it tonight (BJT).

---

By the way, for Zest build, libuv need to be patched if we want to cross-compile for Windows. See:
- https://github.com/AnClark/zyn-fusion-build/commit/245c69be222aff372cd3dcec36af6215d5ecd060

AnClark:
Tested Zest plugin versions just now. (Requires installation via `install-linux.sh`.)

VST2, LV2 and CLAP work, but VST3 failed to be scanned on REAPER. That's so strange.

fundamental:
Hmm, I swear zest loaded previously on wine. Is that consistent with your setup? Apologies for missing the timeslot this weekend, something in my body was out of whack, so I ended up getting very little accomplished in the past few days.

AnClark:
I see. 

For Zest on Wine, maybe that's my environment's issue. I will test it on my native Windows 10.

How are you feeling now? Health is essential, so just have a nice break and take good care of yourself. Hope you'll get better soon.

fundamental:
I'm resuming the review now. First up is fixing the -ldl flag which is still missing in this PR.

fundamental:
LV2 loads the zest UI in ardour, but an error is repeated over and over again indicating a possible regression:
```
Sending key 'state' to UI failed, out of space (needs 20522 bytes)
```

fundamental:
No GUI is displayed for ntk in ardour under LV2 either, which I'm pretty sure is a regression.

falkTX:
> LV2 loads the zest UI in ardour, but an error is repeated over and over again indicating a possible regression:
> 
> ```
> Sending key 'state' to UI failed, out of space (needs 20522 bytes)
> ```

we should update zyn dpf handling to make sure the state is kept dsp-side only. there is no need to inform the gui about state changes from dsp side since there is a whole external OSC layer for it.

falkTX:
> ```
> 
> ```
> 
> 
>     
>   
> 
> Building with FLTK results in a build failure:

this is sorta expected, the cmake side does not handle disabling vst3 ui if it is external only and no embed. something for me to fix on dpf side.



> * 
> 
> * VST2 successfully loads with ntk in ardour (no UI present, but IIRC that's normal for the VST2, ...)

unexpected, with ntk the full UI is meant to be present, since we can use embed

fundamental:
Ok, so reading through the changes everything looks reasonable. It's a bit sad to lose any of the Apple specific bits of CMake, but those haven't been exercised recently. I'm fine with the possible warnings with regards to the state variable, but falktx would be in the right position to declare if there's unforeseen consequences. I would like to have the NTK issues resolved prior to merging because that would absolutely generate user complaints if it is the regression I think it is.

Outside of that and the -ldl/building vst3 when it shouldn't, then this is almost ready to merge. I ought to repeat some of the sanity checks on windows as well, though that is likely constrained to the ubuntu/docker builder + either wine or a windows VM I have kicking around.

AnClark:
NTK works well and can be embedded in REAPER. Does it mean it's specific to other DAWs like Ardour?

falkTX:
The build issue is fixed in https://github.com/DISTRHO/DPF/commit/dc6557a3c935ae4bb68f89d8bc698a805fff788e
Needs a submodule update

falkTX:
@AnClark can you please update the dpf submoduel on this PR?

AnClark:
> @AnClark can you please update the dpf submodule on this PR?

Sure. But many significant changes are still in the `develop` branch, and not synced to `main`. Shall I continue switching to `develop` branch for DPF submodule?

falkTX:
Yes that is fine. The develop branch has new APIs that I want to implement on all formats before merging back to main. Or I just tag them experimental and merge already..
In any case for DPF it is 99% safe most of the time to use the develop branch as-is.

AnClark:
> Yes that is fine. The develop branch has new APIs that I want to implement on all formats before merging back to main. Or I just tag them experimental and merge already.. In any case for DPF it is 99% safe most of the time to use the develop branch as-is.

Thanks. I've synced DPF with the current `develop` branch.

src/Plugin/Chorus/CMakeLists.txt
```diff
@@ -1,40 +1,21 @@
+dpf_add_plugin(ZynChorus
+    TARGETS clap lv2 vst2 vst3
+    FILES_DSP
+        Chorus.cpp
+)
 
-include_directories(${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_SOURCE_DIR}/DPF/distrho .)
+include_directories(.)
```
falkTX, line 7: this should probably use [target_include_directories](https://cmake.org/cmake/help/latest/command/target_include_directories.html)

src/Plugin/Chorus/CMakeLists.txt
```diff
@@ -1,40 +1,21 @@
...
-
-install(TARGETS ZynChorus_lv2 LIBRARY DESTINATION ${PluginLibDir}/lv2/ZynChorus.lv2/)
-install(TARGETS ZynChorus_vst LIBRARY DESTINATION ${PluginLibDir}/vst/)
-
-add_custom_command(TARGET ZynChorus_lv2 POST_BUILD
-    COMMAND ../../lv2-ttl-generator $<TARGET_FILE:ZynChorus_lv2>
-	WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/lv2)
-
-add_dependencies(ZynChorus_lv2 lv2-ttl-generator)
+dpf__determine_vst3_package_architecture(vst3_arch)
```
falkTX, line 11: why is this line needed? the vst3 architecture should already be autodetected by dpf directly.
if that fails, it should be considered a bug and fixed on dpf side.

src/Plugin/ZynAddSubFX/CMakeLists.txt
```diff
@@ -177,52 +89,30 @@ else()
...
-    install(TARGETS ZynAddSubFX_lv2 LIBRARY DESTINATION ${PluginLibDir}/lv2/ZynAddSubFX.lv2/)
-    install(TARGETS ZynAddSubFX_vst LIBRARY DESTINATION ${PluginLibDir}/vst/)
+target_link_libraries(ZynAddSubFX-dsp PRIVATE zynaddsubfx_core zynaddsubfx_gui_stubs
+    ${OS_LIBRARIES} ${LIBLO_LIBRARIES} ${PLATFORM_LIBRARIES})
 
-    add_custom_command(TARGET ZynAddSubFX_lv2 POST_BUILD
-        COMMAND ../../lv2-ttl-generator $<TARGET_FILE:ZynAddSubFX_lv2>
-        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/lv2)
+# Linker diagnostic options
+target_link_libraries(ZynAddSubFX PRIVATE -Wl,--no-undefined,--unresolved-symbols=report-all)
```
falkTX, line 96: this is not compatible with apple clang, but we can sort this after merging as the dpf-ci-workflow PR has build tests for macOS which will tell us any such build issues

src/Plugin/Chorus/CMakeLists.txt
AnClark: Because I will also cross-build for Windows on Linux.

I specify a [toolchain file](https://github.com/AnClark/zyn-fusion-build/blob/new-generation/z/windows-build.cmake) when cross-building on Linux, in [my build script](https://github.com/AnClark/zyn-fusion-build/).

However, when cross-building, DPF CMake script won't get VST3 architecture properly. So I have to manually invoke `dpf__determine_vst3_package_architecture()`.

src/Plugin/Chorus/CMakeLists.txt
AnClark: Did you mean `target_include_directories(ZynChorus ".")`?

src/Plugin/Chorus/CMakeLists.txt
falkTX: right, I handle this on the other PR, but gets tricky to do it automatically.
see https://github.com/DISTRHO/dpf-cmake-action/blob/main/action.yml#L438

I suggest to remove this, and let's try to fix things on dpf side directly.

src/Plugin/Chorus/CMakeLists.txt
falkTX: yes exactly

src/Plugin/ZynAddSubFX/CMakeLists.txt
AnClark: This is used for diagnostics when I did the porting stuff.

If ZynAddSubFX can build and run well with Apple Clang, I can remove this.

src/Plugin/Chorus/CMakeLists.txt
AnClark: Sorry for my mistake.

`dpf__determine_vst3_package_architecture()` is used for generating the VST3 plugin install path. For example:

```cmake
install(TARGETS ZynAddSubFX-vst3 LIBRARY DESTINATION ${PluginLibDir}/vst3/ZynAddSubFX.vst3/Contents/${vst3_arch}/)
```

AnClark: > I want to have the install step be part of dpf too, so eventually this wont be needed.

Great. That should be a better solution.

Maybe you can add some methods, for example, `dpf_install_plugin()`.

JohannesLorenz:
> FYI the port checker test has been known to raise false positives and no changes in this PR should be able to cause a regression that would impact that specific test.

Note about the CI failure: There was a current issue which caused the PortChecker test to sporadically fail. We fixed it now on master. If you rebase on master or merge from master, PortChecker should work for you (or it should correctly highlight port issues that the PR introduces).

JohannesLorenz:
> @JohannesLorenz Seems that CI failure happens when I rebased my repo.


It looks like your branch is based on master commit e9d9fa696d8e928914fd0b83bb29dfab355d5312 (from January). The error is not fixed before ef392d964724236d7c845ff0e5529eed077fbe98 (from May).

AnClark:
Hi @fundamental ! Is it OK to merge this PR? Almost one year passed.

fundamental:
Agreed, it's been too long. I'm going to target merging this this week. Even if regressions do occur this will be a necessary step forward.

fundamental:
As per 5cf300d, the build does not pass unless additional build flags were used. Then with some added testing, a few cstdint header includes were needed (I think the master branch already has those changes). After even more testing for some reason the docker build is now failing, so I'm investigating that.

fundamental:
Oh, there's been quite a few changes. I haven't even gotten to the merging stage. The one failure on the CI seems to be our one flaky test of the port checker. It passes locally so that likely can be disregarded, however this is before building up a merge commit.

fundamental:
Looks like there's some additional changes in this PR that need reverting since it looks like there may be some library naming conflicts with the new folder structure. Additionally it looks like some of the docker scripts stalled out because of the now recursive git submodules (a bit excessive IMO, but it's a minor thing).

fundamental:
VST3 fails to build under mingw, hacking around that problem as well...

fundamental:
And, the LV2 ttl generator fails under mingw as well. Hacking around that problem as well.......

fundamental:
Now it looks like the library name has changed for the windows VST2.dll, breaking another part of the build.

I'm out of time to work on this issue today and while progress has been made I do have to admit it is frustrating to see that this has broken the build in multiple ways even before getting into the merge commit position. This PR is in less of a finished state than I optimistically thought last week.

falkTX:
is the dpf submodule updated to the latest develop version?
I use the dpf cmake stuff quite often (just yesterday I setup https://github.com/DISTRHO/lvgl-template-plugin that uses cmake) and see build going ok for all OSes, so feels weird to me that build is failing..

AnClark:
> Now it looks like the library name has changed for the windows VST2.dll, breaking another part of the build.
> 
> I'm out of time to work on this issue today and while progress has been made I do have to admit it is frustrating to see that this has broken the build in multiple ways even before getting into the merge commit position. This PR is in less of a finished state than I optimistically thought last week.

Could you paste build log here? Maybe I can figure out where failed.

fundamental:
There were a good few different build failures when I was testing last. The below is roughly what I was using (though it looks like some in place hacks to the zynaddsubfx/zynaddsubfx source were lost.

diff on zynaddsubfx/zyn-fusion-build: [build-diff.txt](https://github.com/zynaddsubfx/zynaddsubfx/files/15443896/build-diff.txt) (the docker build script is partially commented out, such that an image can be build and then manually launched to facilitate hacking on the zynaddsubfx/zynaddsubfx repo in place)

@falkTX I was anticipating merging this and then upgrading DPF versions, but if upgrading DPF versions fixes things then that's great. I'm skeptical it will though because most of the compilation in this PR seems to be unconditional which isn't great when one target isn't supposed to be built on all platforms (true at least in the version of DPF used in the PR).

falkTX:
well I now broke things for real in the develop branch, I got rid of the "external ui" support deal in favour of always building a pugl version in stub/dummy mode just to let it deal with all OS event stuff.
as I was using this external UI stuff for webview testing and development, it became clear how much of a pain it is without an OS layer in between.

so now targetting DPF develop so things will break, but I think it is better to do it now than rather.
I want to have a stable DPF release at the end of the year, so reducing the amount of things to maintain is part of that.

if this PR becomes too problematic, maybe we can do it in steps. First updating DPF while keeping the custom cmake stuff, then as 2nd stage let DPF take care of the building.
Though letting DPF do the cmake part sounds easier in theory.

https://github.com/DISTRHO/elements-template-plugin/blob/main/CMakeLists.txt is an example of something built with DPF that uses external libs.

AnClark:
> if this PR becomes too problematic, maybe we can do it in steps. First updating DPF while keeping the custom cmake stuff, then as 2nd stage let DPF take care of the building.

That's a good idea. Seems that still plenty of musicians depend on external UI versions (FLTK, NTK). The CMake implementation with external UI (by the time I submit this PR) is still relevant to ZynAddSubFX.

falkTX:
NTK can do embed, so that still works. but FLTK cant. perhaps the NTK embed patch should be tried to be merged upstream...
worst case the plugin displays a dummy embedded window that needs to be clicked in order to show the real, external ui

fundamental:
Yeah, a fair few users still prefer the old UI, though I will say NTK is somewhat dead compare to FLTK at this point. My main focus though is that this code produces reasonable builds and if there are regressions that we can work towards resolving those. If that means starting with a smaller PR I'm totally open to that.


Also, as a reminder to myself, https://github.com/zynaddsubfx/zyn-fusion-issues/issues/396 should be tested by our test-users after this is merged.

JohannesLorenz:
It's sad that this PR's development has stopped.

The main issues I read here are:

1. CI issues => In a few days or weeks, I will have set up a Github Actions CI that covers MSVC, MinGW, macOS (clang) and Linux. That should be more than enough.
2. It has been suggested to first update the submodule, and then in a later PR try the workflow update.

So here is my suggestion to advance: As soon as #339 (and possibly #340) are solved, we simply update DPF but keep the workflow. We make it compile, maybe test is a bit. Does anyone see issues with this plan?

fundamental:
Related to the CI issues I think I recall seeing that this PR restructured where a number of artifacts were which doesn't break the build per say, but did break packaging scripts. That said, a DPF upgrade is long overdue at this stage.

JohannesLorenz:
Which packaging scripts do you mean?

AnClark:
I remember that this PR is based on the "develop" branch at the time I submitted this PR. Do you plan to apply these changes? I can help Improving this PR to fit with the newest state of the "develop" branch .

AnClark:
Packaging script may also need to modify, because the DPF CMake system has different target filenames.

fundamental:
The packaging scripts over in the zyn-fusion-build repo were broken. I think there might have additionally been complications in terms of renaming lv2/vst versions which would allow the plugin hosts to load zyn, but it would treat it as a new plugin rather than a version change on the old one. IMO those changes ought to be relatively small.

We certainly do want to update DPF and @JohannesLorenz has been making some nice improvements to our CI which should help keep things in sync after the update. There may be a bit of pending work that I still have to finish off for updating mruby-zest-build with a newer version of PUGL, but a PoC branch for that work was made this year.

AnClark:
I've resolved conflicts reported in this PR.

CI should works well, as the CMake workflow is almost the same as before. This PR mainly replaces those manual build commands to the more convenient `dpf_add_plugin()` call.

But it seems that CI fails on LLD and Zyn-Fusion. There's still something to do.

PR #223, opened 2023-03-18
Fixup DynFilter presets after corruption in 2017

JohannesLorenz:
This is likely a fixup of b650636e306237397ef106b86fc2aaa35ec5182b, where those params have been turned into mostly unused dummy params, but the DynFilter code which uses them has been forgotten to adapt.

So this code fixes the DynFilter presets, so they sound like in 2017 again.

Additionally, this code also removes the dummy variables, which have no use as class variables since 2017.


JohannesLorenz:
This only fails for port checker, but port checker passes locally for me (and this code does not change the ports).

So IMO it can be merged.

fundamental:
Merging as this is consistent with a previous conversation in a development chat. Thanks for spotting this oversight in the previous DynFilter work.

PR #224, opened 2023-03-26
Sympathetic: Make "Generic" a preset

JohannesLorenz:
As discussed on Discord.

### DIFF
--- a/src/Effects/Sympathetic.cpp
+++ b/src/Effects/Sympathetic.cpp
@@ -41,25 +41,32 @@ rtosc::Ports Sympathetic::ports = {
                   else
                       d.reply(d.loc, "i", o->Ppreset);
                   rEnd},
-    rEffParVol(rDefault(127), rPresets(100, 80, 100, 90)),
-    rEffParPan(rDefault(64)),
+    {"presetOfVolume:", rProp(internal), 0,
+        rBegin; (void)msg;
+        rObject *o = (rObject*)d.obj;
+        d.reply(d.loc, "i", o->Ppreset + (16 * o->insertion));
+        rEnd},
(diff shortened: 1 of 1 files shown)


friedolino78:
I am not quite sure anymore about the idea to make "generic" a preset.
Maybe we should think again about the preset handling in general.
My idea is that i can choose a preset to have the params set. If I change a value, it is no more the preset. But the gui still shows the preset name. Therefor I wanted to have a way to say "no preset". Making this "no preset" a preset gives me a bit of logical discomfort.

JohannesLorenz:
You mean you want the GUI to always display "no preset" as soon as you changed a parameter? Then this could be a GUI-only fix.

One reason to still display the preset might be "It's still a Hall reverb, just maybe with 1 or 2 changed parameters" - if the preset is not displayed anymore, the user won't know if it's a Room/Hall/Tunnel reverb.

One compromise could also be to just let the GUI gray-out the preset as soon as one parameter is changed. Or display all derivating parameters (which could be done since we know the default values for each preset).

fundamental:
Shifting how presets are shown sounds interesting. It might be a bit of a messy implementation, but I'd expect users would like it. Per the code, making updates only on parameter differences sounds good to me. Were the removal of some of the rDefault() statements intentional?

JohannesLorenz:
> Per the code, making updates only on parameter differences sounds good to me.

Which kind of updates do you mean?

> Were the removal of some of the rDefault() statements intentional?

Yes. The idea was that the presets have different default values for those ports, so there is no "one default" value that matches "most of the presets except a few".

fundamental:
> Which kind of updates ...

These updates:
```
            if(Punison_frequency_spread != value)
```

friedolino78 (approved):
looks good to me

PR #225, opened 2023-04-01
Release preparation: updated authors and copyright

pgervais:


fundamental:
Thanks for the resync of the authorship file :)

PR #229, opened 2023-04-14
sync ADSR to ENV points in free mode

friedolino78:


fundamental:
Notes from development chat on this PR. There previously were issues with feedback loops generated with the /damage calls in the fusion GUI (not the FLTK GUI if my memory is right). So, something needs to be done to the fusion GUI to ensure it doesn't overfetch/overupdate parameters in the envelope visualizations.

I also think that there _might_ have been some need for sanity checks on the Penvsustain and Penvpoints values for very short envelopes.

JohannesLorenz:
Note about the CI failure: There was a current issue which caused the PortChecker test to sporadically fail. We fixed it now on master. If you rebase on master or merge from master, PortChecker should work for you (or it should correctly highlight port issues that the PR introduces).

PR #230, opened 2023-04-16
OscilGen: Skip no-ops for custom base funcs

JohannesLorenz:
This skips an unnecessary re-computation of OscilGen's `basefuncFFTfreqs`.

Before this patch, not only runtime is wasted, but the redundant FFT can slightly modify the base func FFT freqs, which can cause issues when comparing savefiles.


fundamental:
Agreed with the changes as it's consistent with our earlier conversation.

Merged :+1:

PR #231, opened 2023-04-21
Fix issue with ADDsynth part going silent in Zynthian when Resonance enabled

polluxsynth:
This probably wouldn't show up on any other platform, as it is based on something sending CC#78 (Resonance Bandwidth) to Zynaddsubfx (without updating Resonance Bandwidth Depth first), but Zynthian does, and when it happens it triggers a bug which I believe causes disruption of the audio when Resonance is enabled due to the relbw struct member being uninitialized. I didn't debug it further than this, basically just copying the code from the way Resonance Center is handled. Testing adjusting Resonance Center and Resonance Bandwidth using the Zynthian built-in GUI now behaves in a sane manner so it seems to be the correct fix.

While debugging I also found a discrepancy in the way incoming CC messages for Resonance Center and Resonance Bandwidth are handled, again, essentially copying code from one method to the other. I haven't seen any ill effects, it just looked like both should be handled the same way. Since it's not the same issue, I've submitted it as a separate commit.

### DIFF
--- a/src/Misc/Part.cpp
+++ b/src/Misc/Part.cpp
@@ -847,8 +847,13 @@ void Part::SetController(unsigned int type, int par)
             break;
         case C_resonance_bandwidth:
             ctl.setresonancebw(par);
-            kit[0].adpars->GlobalPar.Reson->
-            sendcontroller(C_resonance_bandwidth, ctl.resonancebandwidth.relbw);
+            for(int item = 0; item < NUM_KIT_ITEMS; ++item) {
+                if(kit[item].adpars == NULL)
+                    continue;
+                kit[item].adpars->GlobalPar.Reson->
+                sendcontroller(C_resonance_bandwidth,
+                               ctl.resonancebandwidth.relbw);
+            }
             break;
     }
 }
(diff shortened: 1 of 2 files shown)


polluxsynth:
I see that the PortChecker test has failed in CI. I'll have to have a look at it and revisit the patch.

fundamental:
I wouldn't worry about the portchecker test. It doesn't look like this code should impact that particular test, so it must be a false positive.

polluxsynth:
Ok good. If you're not worrying then I won't either. :-)

In fact, when I do 'make test' on my local machine here, I don't get any failures in the PortChecker.

BTW, is there any way for me to access the detailed log from the CI reportedly in /home/runner/work/zynaddsubfx/zynaddsubfx/build/Testing/Temporary/LastTest.log ? I'm still curious as to what has failed.

fundamental:
It looks like we'd have to change https://github.com/zynaddsubfx/zynaddsubfx/blob/master/.github/workflows/ccpp.yml#L27 to run `ctest --output-on-failure` to see that output.

fundamental:
I'm a-ok to merge the included behavior change and glad to hear it resolves your issue. Merged :+1:

PR #233, opened 2023-04-22
Don't write invalid values to XML savefiles

JohannesLorenz:
* Part.cpp: keylimit is 100 in some old savefiles
* ADnoteParameters.cpp: unison size was 0 or >50 while developing the zest UI
* FilterParams.cpp: basefreq was sometimes slightly lower than "min" (rounding errors)

### DIFF
--- a/src/Misc/Part.cpp
+++ b/src/Misc/Part.cpp
@@ -1246,7 +1246,8 @@ void Part::add2XML(XMLwrapper& xml)
     xml.addparbool("note_on", Pnoteon);
     xml.addparbool("poly_mode", Ppolymode);
     xml.addpar("legato_mode", Plegatomode);
-    xml.addpar("key_limit", Pkeylimit);
+    unsigned char keylimit_max = std::atoi(ports["Pkeylimit"]->meta()["max"]);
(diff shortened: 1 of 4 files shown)


src/Misc/Part.cpp
```diff
@@ -1246,7 +1246,8 @@ void Part::add2XML(XMLwrapper& xml)
     xml.addparbool("note_on", Pnoteon);
     xml.addparbool("poly_mode", Ppolymode);
     xml.addpar("legato_mode", Plegatomode);
-    xml.addpar("key_limit", Pkeylimit);
```
hselasky, line 1249: atoi() parses an integer. Are there any problems about that?
Should you use "int" ??

fundamental: Ah, you've got a build where POLYPHONY is more than 127, right? In that case it should be easy to change `unsigned char keylimit_max` to `int keylimit_max`

JohannesLorenz: Sorry, this was already fixed in #235 , but this time at loading time, not at saving time. So this PR is deprecated.

Also, in #235 , I indeed used "int" here, so it should be fine.

PR #236, opened 2023-04-24
Fixed compilation of zynaddsubfx-ext-gui on MacOS

pgervais:
Original issue: `make zynaddsubfx-ext-gui` failed to compile because it couldn't find the png library. The -lpng flag is generated by fltk-config, which apparently assumes that it's installed system-wide and doesn't require a -L flag. This assumption is incorrect when libpng has been installed through Homebrew.

The fix here is to pass the homebrew lib/ directory to the linker, when homebrew is installed.

Test system: Macbook Air M2 (arm64) running MacOS Ventura 13.3.1, using Homebrew to install the required dependencies. Compilation is using Apple's Clang from XCode command-line tools. 

With this commit, the **FLTK standalone version** compiles and runs (using jack as a backend, though I haven't been able to hear any sound yet). 

**Plugin versions of Zyn don't compile**, apparently because of a function not yet implemented in DPF. Once https://github.com/zynaddsubfx/zynaddsubfx/pull/221 is merged I'll take another stab at it.

### DIFF
--- a/cmake/FindHOMEBREW.cmake
+++ b/cmake/FindHOMEBREW.cmake
@@ -1,0 +1,10 @@
+#Find Homebrew
+
+if ("Darwin" STREQUAL ${CMAKE_SYSTEM_NAME})
+  if(DEFINED ENV{HOMEBREW_PREFIX}) 
+    set(HOMEBREW_LIB_DIR "$ENV{HOMEBREW_PREFIX}/lib")
+  endif()
+endif()
+
+include(FindPackageHandleStandardArgs)
+FIND_PACKAGE_HANDLE_STANDARD_ARGS(HOMEBREW DEFAULT_MSG HOMEBREW_LIB_DIR)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -18,6 +18,7 @@ find_package(Alsa)
 find_package(Sndio)
 find_package(FLTK)
 find_package(OpenGL) #for FLTK
+find_package(HOMEBREW)  #for libpng and libjpeg (for fltk on MacOS)
 # lash
 if(PKG_CONFIG_FOUND AND NOT (${CMAKE_SYSTEM_NAME} STREQUAL "Windows"))
     message("Looking For pkg config modules")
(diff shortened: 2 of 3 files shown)


falkTX:
Is there no `${FLTK_LIBRARY_DIRS}` ? that seems like it would be the better approach

pgervais:
In the current state of the CMakeList.txt file, the ldflags are obtained using fltk-config, which does not provide any information regarding libpng and libjpeg other than the -lpng and -ljpeg flags, which is why I went for this indirect solution.

That said, looking a bit closely at the history it appears that everything related to fltk in src/CMakeList.txt has been added a long while ago, and the FindFTLK.cmake file has evolved a lot since then. In particular it looks like it may provide access to the values we need, as you suggested. I'll dig deeper.

falkTX:
my suspicion is that the same dir where fltk is located, is where libpng and other libs will be located too.
so I dont think we need all specific libdirs, just the fltk one.
this all assumes fltk libdir is available anyway

pgervais:
Well... if it were that simple there wouldn't be a need for this change :) 

libpng and libfltk are both symlinked in /opt/homebrew/lib, but the problem is that fltk-config returns a more specific path, which doesn't contain libpng: /opt/homebrew/Cellar/fltk/<version>/lib. Maybe the default FindFLTK.cmake will not rely on fltk-config, and will instead return the common directory. In that case that would solve the problem.

fundamental:
Well I wasn't expecting to see a MacOS fix given that https://github.com/zynaddsubfx/zynaddsubfx/pull/221 is likely to break some MacOS functionality. I'll give this a second read through, but it seems like a sane enough workaround given that `fltk-config --ldflags` must be inaccurate on that platform.

pgervais:
Let's call that a warmup for me :) Now that I have access to a MacOS machine I should be able to devote some time maintaining the MacOS port. That said, if you think there is no demand for MacOS anymore and the project should stop supporting it, happy to withdraw that PR.

pgervais:
I will close this PR in favor of the other one I linked above.

PR #237, opened 2023-04-29
Implement/use transmitMsgGui

JohannesLorenz:
@fundamental Can you please review this? The implementatin of `transmitMsgGui*` is exactly like we discussed in Discord, so you don't need to check this. But it would be nice if you can check if I replaced `transmitMsg*` by `transmitMsgGui*` in the right places and did not forget elsewhere - I wrote down the justification below in the commit description. Thanks on advance.

--- commit description below ---

This fixes an issue only related to in-process remotes, aka "GUI".

`MwDataObj::reply` uses `MiddleWareImpl::sendToCurrentRemote`, which sometimes uses `MiddleWareImpl::last_url` as the reply URL. However, the latter variable is only being set in `handler_function`, i.e. in the liblo callback, so in case the GUI was last, `last_url` has not been updated, and `MwDataObj::reply` sends to "nowhere".

This commit provides `transmitMsgGui*` functions, which are like the `transmitMsg*` functions, except that they set `last_url` accordingly. All usages of `transmitMsg*` shall be transferred to `transmitMsgGui` if being done in GUI context.

Where did we replace use of `transmitMsg*` to `transmitMsgGui*`?

* UI/: everywhere, obviously
* Tests/: everywhere, because the tests are always in-process connected
* PresetExtractor.cpp: nowhere, because all calls are in callbacks which were already preceeded by a call to either `transmitMsg*` or `transmitMsgGui*`, which has already set `last_url` correctly
* MiddleWare: nowhere, because all calls are either in the `handler_function`, or they are in callbacks like `loadPart`, which were already preceeded by a call to either `transmitMsg*` or `transmitMsgGui*` which has already set `last_url` correctly
* main.cpp: Nowhere, because a reply does not make sense here

### DIFF
--- a/src/Misc/MiddleWare.cpp
+++ b/src/Misc/MiddleWare.cpp
@@ -2547,6 +2547,36 @@ void MiddleWare::transmitMsg_va(const char *path, const char *args, va_list va)
         fprintf(stderr, "Error in transmitMsg(va)n");
 }
 
+void MiddleWare::transmitMsgGui(const char *msg)
+{
+    if(activeUrl() != "GUI") {
+        transmitMsg("/echo", "ss", "OSC_URL", "GUI");
+        activeUrl("GUI");
+    }
+    transmitMsg(msg);
+}
+
+void MiddleWare::transmitMsgGui(const char *path, const char *args, ...)
+{
+    char buffer[1024];
+    va_list va;
+    va_start(va,args);
+    if(rtosc_vmessage(buffer,1024,path,args,va))
+        transmitMsgGui(buffer);
+    else
+        fprintf(stderr, "Error in transmitMsgGui(...)\n");
+    va_end(va);
+}
+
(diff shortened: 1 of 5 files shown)


fundamental:
The tests generally wouldn't need this functionality and in fact I'd argue if transmitMsgGui was implemented in the way I'd expect the tests would fail. As proposed transmitMsgGui sets the active URL (things like snoop ports), but it does not queue an update for the currentURL via sending the OSC_URL update message to the backend and waiting for it to appear in order later on.

Somewhat non-intuitively guimain.cpp would not use transmitMsgGui as that is the code used to make the fltk/ntk UI run out-of-process.

JohannesLorenz:
@fundamental I tried to fix the mentioned issues and force-pushed a correct version. Can you please check the new code and commit message? Thanks.

fundamental:
The modified commit appears to address all issues discussed here and on the dev chat. From my reading the commit message is accurate to the situation in zyn. Merged :+1:

PR #238, opened 2023-05-01
Make cmake error out when ntk-fluid is not found

pgervais:
It's a simple change that will make compiling zynaddsubfx with the ntk UI easier for newcomers.
I checked that this PR doesn't break the windows and mac builds.

-- Commit message:
This fixes several confusing issues that happen when both fltk and ntk are installed. The default is to compile using ntk, but the fltk installs somehow shadows ntk-fluid, leading to fl files not being converted to cxx. This only shows at compilation time, very far from the original issue.
This changes does two things:
- prevent fltk fluid from shadowing ntk-fluid
- make the cmake invocation idempotent: when the FLTK_FLUID_EXECUTABLE is not unset, the first cmake invocation fails, but the second one goes through because FLKT_FLUID_EXECUTABLE is set to the default value 'fluid'.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -479,9 +479,13 @@ if(FltkGui)
 endif()
 
 if(NtkGui)
+  unset (FLTK_FLUID_EXECUTABLE)
+  find_program(FLTK_FLUID_EXECUTABLE ntk-fluid)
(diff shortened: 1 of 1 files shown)


fundamental:
It sounds like this was a painful one to debug.

Thanks for the enhancement. Merged :+1:

PR #240, opened 2023-05-01
Fixed compilation of zynaddsubfx-ext-gui on MacOS

pgervais:
Removed the 'ugly workaround' which doesn't seem to be necessary anymore with recent versions of cmake, and added missing link directory flags that were missing.


pgervais:
@falkTX @fundamental Let me know what you think. I am aware that #221 damages MacOS's support, but it looks like the present change is orthogonal: it's about fixing the standalone gui compilation.

fundamental:
This feels like the sort of change that's going to look fine but break things on a single linux distro :p

One minor change request, can you update your commit message to conform to the style listed in https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html specifically capitalization, verb tenses, and line widths. I'm mostly bringing it up to avoid the line wrap issues on viewing logs in the command line.

fundamental:
Thanks for the fixup. Merged.

PR #242, opened 2023-05-03
Fix individual synth panning direction (#376)

polluxsynth:
Fix panning direction for the individual synths as noted in the issue.

Also fix the ADnote test which actually is panning dependent. Simple fix, just use the R channel from the synth output instead of the L channel (this also ensures a certain amount of backwards traceability rather than extracting new values to compare with in the applicable tests).

### DIFF
--- a/src/Synth/ADnote.cpp
+++ b/src/Synth/ADnote.cpp
@@ -1837,9 +1837,9 @@ int ADnote::noteout(float *outl, float *outr)
             if(stereo)
                 for(int i = 0; i < synth.buffersize; ++i) { //stereo
                     outl[i] += tmpwavel[i] * NoteVoicePar[nvoice].Volume
-                               * NoteVoicePar[nvoice].Panning * 2.0f;
-                    outr[i] += tmpwaver[i] * NoteVoicePar[nvoice].Volume
                                * (1.0f - NoteVoicePar[nvoice].Panning) * 2.0f;
+                    outr[i] += tmpwaver[i] * NoteVoicePar[nvoice].Volume
+                               * NoteVoicePar[nvoice].Panning * 2.0f;
                 }
             else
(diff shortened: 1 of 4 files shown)


hselasky:
Hi,

Your change looks correct, but looking at the code I wonder if we should make a "gain" class in Master.h, providing a float gain[2] . Then you don't have to compute the gain inside for-loops over and over again? Not sure if all compilers can optimise such. One float multiplication instead of two?

--HPS

polluxsynth:
That's a good point. Not sure if it's worth having a class for it though. Especially since the calculations for whatever reason are different for ADsynth compared to the others. A simple precalculation of lgain and rgain would keep all but the actual sample calculation out of the loop. I'll have a look at it.

hselasky:
@polluxsynth : Something like that. Maybe some common function can do it. I think you got it!

fundamental:
Thanks for the fixes, there's been a number of mistakes in panning over the years, so hopefully this resolves some of the issues. Merged :+1:

polluxsynth:
Thanks @fundamental!

Addressing @hselasky's concern, one thing I can't get my head around is why the right and left pan in ADDsynth is calculated as **Panning * 2.0f** and **1.0f - NoteVoicePar[nvoice].Panning * 2.0f**, respectively, whereas for SUBsynth and PADsynth they are the more comprehensible **Panning** and **1.0f - Panning** . In all cases, Panning is in the range 0..1 , being either a random float in the range 0..1 or the actual panning parameter divided by 127 or 128.

So, **Panning * 2.0f** results in a value range of 0..2 (for the right channel), whereas **1.0f - NoteVoicePar[nvoice].Panning * 2.0f** results in a value range of +1..-1 (for the left channel). Armed with this newfound reallization, I have yet to listen critically to the result though.

PR #243, opened 2023-05-04
Add DPF-based CI workflow

falkTX:
This imports the magic over at https://github.com/DISTRHO/dpf-cmake-action and https://github.com/DISTRHO/PawPaw/ for dealing with static builds on Linux, macOS and Windows.
Basically this builds most common opensource libs statically, hides all symbols, tweaks builds for self-contained use etc etc and generates useful general use binaries at the end.
For zynaddsubfx this means zlib, mxml, fftw3 are built and linked to statically.
There are many other libs in there, like libsndfile, cairo and glib, but those are not needed here.

No fltk/ntk UI support for now, it is something I can add into PawPaw later on. From my experience building fltk is not complicated anyway.

Note this doesn't really work for now, mainly because it requires #221 to be merged first.
But can already be useful to detect a few build issues beforehand.


hselasky:
Does this mean you will have per-built binaries for some platforms?

falkTX:
soon soon, yes. at least for linux, macos and windows.

PR #247, opened 2023-05-26
Remove unsupported flag on MacOS

pgervais:
The --no-as-needed flag is not supported by Clang. Removed it so that the compilation of tests can go through.


fundamental:
I suspect we can remove the whole "-Wl,--no-as-needed -lpthread" chunk as glancing at the git logs it was a workaround for a gcc 4.8 bug, which is likely no longer relevant.

fundamental:
Thanks for raising the issue. The alternative patch has been merged.

pgervais:
Thanks. I cherry-picked the fix into the release-3.0.7 branch.

PR #248, opened 2023-06-10
Bump rtosc's version

pgervais:
This moves rtosc past the port-checker's refactor commit, which will fix the windows build.


pgervais:
Aborting this PR it's the wrong fix.

PR #256, opened 2023-07-21
Hysteresis

friedolino78:
this PR adds a new effect Hysteresis that should implement the Jiles-Atherton model of magnetic hysteresis

### DIFF
--- a/src/Effects/CMakeLists.txt
+++ b/src/Effects/CMakeLists.txt
@@ -12,5 +12,7 @@ set(zynaddsubfx_effect_SRCS
 	Effects/Phaser.cpp
 	Effects/Reverb.cpp
 	Effects/Sympathetic.cpp
+	Effects/Hysteresis.cpp
+	Effects/JilesAtherton.cpp
     PARENT_SCOPE
 )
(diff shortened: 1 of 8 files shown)


friedolino78:
gui PR: https://github.com/mruby-zest/mruby-zest-build/pull/106

fundamental:
It looks like you're experimenting around with different methods. Locally I can use the first approach (with modifications) to get to something that does work and is symmetric, though I can see how simple parameter changes can get you to the asymmetry issue you've noted this effect can have.


The below are what I get from implementing the Static Jiles-Atherton Model from Dynamic Jiles-Atherton Model for Determining the Magnetic Power Loss at High Frequency in Permanent Magnet Machines (Though I don't know what they're talking about with the δ variable as the method appears to fail to converge unless δ==+1.0)

magnetic-field
xyplot


And a bonus saw wave:

magnetic-field-saw

It looks like your second stab at the problem is still a WIP, so I'll let you continue to hack away at the system of equations.

friedolino78:
I got a little bit lost with these algorithms. I want exactly what you did.

fundamental:
Rename the linked .txt file to html and you should be able to see the notebook that was able to generate the figures. As mentioned in the document it's not something suitable for deployment to a full algorithm. For that use case you'd want something where you had more of a closed form solution as opposed to the crude iterative numeric method I cobbled together.

[2023-07-29-hysteresis.jl.txt](https://github.com/zynaddsubfx/zynaddsubfx/files/12307548/2023-07-29-hysteresis.jl.txt)

PR #258, opened 2023-08-05
fixed cstdint issue in msys2/mingw64 windows compilation

rabidmachine9:
Had to include <cstdint> in a few more files for compilation to be successful

### DIFF
--- a/src/Params/ADnoteParameters.h
+++ b/src/Params/ADnoteParameters.h
@@ -16,6 +16,7 @@
 
 #include "../globals.h"
 #include "PresetsArray.h"
+#include <cstdint>
 
 namespace zyn {
 
--- a/src/Params/Controller.h
+++ b/src/Params/Controller.h
@@ -17,6 +17,7 @@
 
 #include <stdint.h>
 #include "../globals.h"
+#include <cstdint>
 
 namespace zyn {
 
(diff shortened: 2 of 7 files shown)


fundamental:
To confirm, the `#include <stdint.h>` in SubnoteParameters.h, FilterParams.h, and Controller.h was insufficient and the `#include <cstdint>` change was required to build those files?

rabidmachine9:
Sorry for any miscommunication,

I show that you just added #include <stdint.h> to just one file FilterParams.h and that was not enough.
I knew that since I had tried it before and it found another conflict down the way, since I thought it was just a rabbit whole where I would just keep editing files and new errors would occur I decided to ask a question first. Your comment game me some assurance that this was a minor issue that can be resolved this way so I searched "Params" folder for any occurrence of int64_t and I added the headers in each and every file that used it. 
I didn't go the edit one file and try to compile way since it seemed a dead end and compilation takes a good amount of time. 
The reason I used cstdint instead of cstdint.h is probably because cstdint is what is mentioned in the error output so I tried that and it worked. Also since my knowledge of C++ is limited I had to ask ChatGPT for the difference between the 2 headers and I got this response: "there is no standard header file named cstdint.h". I decided to change your line of code just for the sake of consistency and I meant no disrespect. 
Finally I decided to make the pool req in order to get me started with the project and maybe help some others newbies with compilation ?

Thanks for all the support and have a good one :)

fundamental:
Sounds reasonable enough to me. The `stdint.h` vs `cstdint` different shouldn't cause any issues elsewhere, so thanks for resolving the build issue :)

Merged :+1:

Issue #261, opened 2023-10-08
FM is corrupted


JohannesLorenz:
FM seems to fail in the "sine modulates sine" case in many scenarios. Some may fall in the aliasing area, but the most remarkable scenario is this:

1. Start zyn with SR 192000 (error occurs with all SRs, but let's exclude aliasing issues)
2. Use a fresh AD voice
3. Set modulator type to FM
4. Set FM volume to 120
5. Activate AMP env
6. Set stretch to 127
7. Play the lowest note on the internal "vK" (virtual keyboard)

After 0.3 seconds, you will hear a sudden transition which can not be caused by aliasing. This must be investigated, possibly with debug printfs at the time where the sound changes.

Also notable, if you switch to PM, the issue is gone. Except for initial parameter,s PM and FM share almost the same modulation code, except for the normalization phase: https://github.com/zynaddsubfx/zynaddsubfx/blob/806674849b276a560d2996f25ba0ddbfe50838ef/src/Synth/ADnote.cpp#L1524-L1545

friedolino78:
@JohannesLorenz 
please check if this bug is fixed with:
https://github.com/zynaddsubfx/zynaddsubfx/pull/264

JohannesLorenz:
> I don't remember. feel free to use it if needed

This must be the reason: "I did some spring cleanup by closing PRs that have become obsolete.". Not sure why you thought it was obsolete. I will reopen it for now.

PR #264, opened 2023-11-03
fix limiting FMmodfreqlo

friedolino78:
this fix eliminates some spectral discontinuities when raising the fm amount.

### DIFF
--- a/src/Params/EnvelopeParams.cpp
+++ b/src/Params/EnvelopeParams.cpp
@@ -539,7 +539,7 @@ void EnvelopeParams::add2XML(XMLwrapper& xml)
 }
 
 float EnvelopeParams::env_dB2rap(float db) {
-    return (powf(10.0f, db / 20.0f) - 0.01)/.99f;
+    return std::max(powf(10.0f, db / 20.0f) - 0.01,0.0)/.99f;
 }
 
 float EnvelopeParams::env_rap2dB(float rap) {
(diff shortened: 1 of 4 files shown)


fundamental:
The first commit feels plausibly correct, the second one feels like it would create possible segfaults (in my mind at least). Removing the fmodf() means that tw[i] shifts from being bounded 0..oscillator-size, to being unbounded. Without that constraint then the integer version (the 'high' variable) is again unbounded, so is the code correct with that bound removed?

Also, since there was a flurry of discussion around the FM modulation, let me know if this PR became out-of-date.

friedolino78:
this PR is the fix for 
https://github.com/zynaddsubfx/zynaddsubfx/issues/261

the unbounded value goes through this chain:
fmold -> tw[i] -> FMmodfreqhi -> carposhi
where some other values are added,
and is bounded then to the correct boundaries.
carposhi &= (synth.oscilsize - 1);

the bounding of fmold, i want to remove leads to discontinuity in the modulator -> frequency jump

fundamental:
ah, as long as carposhi is bounded, then I think this is fine (assuming it improves the FM issues that were being experienced)

JohannesLorenz:
That bug that I mention here is already fixed by my Envelope fix (which I thought was part of Friedonlino's PRs, but I cannot find it right now). You remember that fix with the weird constants 0.01 and 0.99, where I also asked Paul Nasca (no reply yet)?

friedolino78:
I think and hope the latest commit fixes the fmamp envelope problem of #261.
it seems to be a typo at the definition of MIN_ENVELOPE_DB that affects only amp envelopes at values around 0.

friedolino78:
i removed the special treatment for small values in amp envelopes because they start with 0 even without it.

JohannesLorenz:
@friedolino78 Can you please rebase on current master? Should be easy.

friedolino78:
> @friedolino78 Can you please rebase on current master? Should be easy.

done

PR #265, opened 2023-11-05
Fmaa multiscale

friedolino78:
this commit adds a function to calculate a windows sinc kernel.
this is used as antialiasing filter in addsynth for the standard oscillator if the AA button in voice list
it effectively reduces audible antialiasing which otherwise occurs if a note is bent to higher frequencies by ENV, LFO or pitchbend.

In FM it is always active because there is such frequency bending the standard use case.


src/Params/ADnoteParameters.cpp
```diff
@@ -720,6 +724,7 @@ ADnoteGlobalParam::~ADnoteGlobalParam()
     delete FilterEnvelope;
     delete FilterLfo;
     delete Reson;
+    delete wskernel;
```
JohannesLorenz, line 727:
```suggestion
    delete [] wskernel;
```

PR #266, opened 2023-11-05
add Oscillator Sync as new modulation mode

friedolino78:
noting more to say.

to use it you have to  zynaddsubfx --dump-json-schema=mruby-zest-build/src/osc-bridge/schema/test.json


JohannesLorenz:
Thanks for the submission.

However, it is missing an FLTK implementation. If we do not take this into FLTK, then all following modulators will not work in FLTK, because the numbers (enum) of the modulators between FLTK and Fusion will differ. Can you please add it to FLTK?

JohannesLorenz:
I added a mini commit which adds the FLTK menu entry

friedolino78:
> I added a mini commit which adds the FLTK menu entry

thank you.

PR #267, opened 2023-11-17
Fmaa sync

friedolino78:
this PR contains the FM fixes the improved antialiasing for FM and standard and the Oscillator sync for all FM modes
GUI support:

### DIFF
--- a/src/Params/ADnoteParameters.cpp
+++ b/src/Params/ADnoteParameters.cpp
@@ -176,7 +176,7 @@ static const Ports voicePorts = {
 
     //Modulator Stuff
     rOption(PFMEnabled, rShort("mode"), rOptions(none, mix, ring, am, phase,
-                frequency, pulse, sync, syncrm), rLinear(0,127), rDefault(none), "Modulator mode"),
+                frequency, pulse), rLinear(0,127), rDefault(none), "Modulator mode"),
     rParamI(PFMVoice,                   rShort("voice"), rDefault(-1),
         "Modulator Oscillator Selection"),
     rParamF(FMvolume,                   rShort("vol."),  rLinear(0.0, 100.0),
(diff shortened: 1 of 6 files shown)


src/Params/ADnoteParameters.cpp
```diff
@@ -720,6 +726,7 @@ ADnoteGlobalParam::~ADnoteGlobalParam()
     delete FilterEnvelope;
     delete FilterLfo;
     delete Reson;
+    delete wskernel;
```
JohannesLorenz, line 729: Valgrind showed an error here. The `new[]` must be matched with `delete[]`, not with `delete`.
```suggestion
    delete [] wskernel;
```

src/Params/ADnoteParameters.cpp
```diff
@@ -175,8 +175,8 @@ static const Ports voicePorts = {
 
 
     //Modulator Stuff
-    rOption(PFMEnabled, rShort("mode"), rOptions(none, mix, ring, phase,
-                frequency, pulse), rLinear(0,127), rDefault(none), "Modulator mode"),
+    rOption(PFMEnabled, rShort("mode"), rOptions(none, mix, ring, am, PM,
+                FM, PWM), rLinear(0,127), rDefault(none), "Modulator mode"),
```
JohannesLorenz, line 179: This somehow does not match the FLTK items, right? I would expect something like `none, mix, ring, PM, FM, PWM, FMAA-SYNC`.

friedolino78: > This somehow does not match the FLTK items, right? I would expect something like `none, mix, ring, PM, FM, PWM, FMAA-SYNC`.

no, 
the sync is enabled with PsyncEnabled. it can be combined with every modulation mode
the AA is enabled with the existing AAenabled port.
only new thing is AM mode which is a variation of ring

friedolino78:
order of modulations is not backward compatible to old presets

friedolino78:
AM and SYNC stuff removed via chunk staging and rebased to current master.

fundamental:
I see a few tests are failing. Should the code be expected to yield identical values to the old code with anti-aliasing switched off?

JohannesLorenz:
I think this should be moved from "In Review" to "Rework"?

PR #268, opened 2023-11-20
OscilGen's Resonance can be const

JohannesLorenz:
Minimal PR, should not cause any problems. Will merge in a few days if no complaints.


fundamental:
Merged :+1:

PR #269, opened 2023-11-22
add negative Combfiltertypes

friedolino78:
the combfilters with negative gain coefficients are quite handy for example to reduce even harmonics


fundamental:
Everything seems to line up. Thanks for the enhancement and merged :+1:

PR #273, opened 2024-01-03
add Korg35 Filter

friedolino78:
gui: https://github.com/mruby-zest/mruby-zest-build/pull/112

### DIFF
--- a/src/DSP/CMakeLists.txt
+++ b/src/DSP/CMakeLists.txt
@@ -5,8 +5,10 @@ set(zynaddsubfx_dsp_SRCS
     DSP/FormantFilter.cpp
     DSP/SVFilter.cpp
     DSP/MoogFilter.cpp
+    DSP/SallenKeyFilter.cpp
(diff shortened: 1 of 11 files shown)


fundamental:
As per the CI, not all the files needed to run this branch are checked in.

fundamental:
The formulation of the filter reminds me of our state variable filter which is composed out of multiple types of filters simultaneously. I suspect the difference that we were able to hear in the audio example was mostly due to the radically different Q mapping levels (there seems to be a non-linear change from the existing filter sets and the max Q is substantially lower than other filters). I do enjoy having a non-linearity in the feedback path, but that tanh approximation is pretty far away from tanh. Is that intended?

As per the claims of faster modulation I tried strong envelopes and strong LFOs and saw similar results to our existing filters, but perhaps I did not try an extreme enough case to replicate the behavior on your end.

friedolino78:
i guess we leave this one out then

PR #274, opened 2024-01-05
new chorus mode - Roland Ensemble

friedolino78:
this adds a new chorus mode ensemble, that attempts to realize the three-phase chorus 
once established by the Roland RS202 string synthesizer
https://www.soundonsound.com/reviews/roland-rs202-multivox-mx202-retrozone
https://www.florian-anwander.de/roland_string_choruses/

more info about 3 phase choruses:
http://jhaible.com/legacy/triple_chorus/triple_chorus.html

gui at: https://github.com/mruby-zest/mruby-zest-build/pull/113

# Description

(Write your PR description here)

# Checks

## Human checks (please check those!):

- [x] 01. Could parts of this PR be split into another PR (while keeping the PRs "atomic")?
- [x] 02. Does this PR depend on another PR or does it block another PR?
- [ ] 03. Does this PR solve a known issue?
- [x] 04. Is there sufficient user documentation in the `doc/` folder?
- [x] 05. Is the code realtime safe? (e.g. no `new`/`delete`, no other syscalls like `sleep`/file handling/mutexes, no prints in non-error code pathways, ...)
- [x] 06. Is aliasing avoided?
- [x] 07. Are clicks/pops/discontinuities avoided?
- [x] 08. Are any new parameters well mapped over their provided range?
- [x] 09. Is divide by zero impossible?
- [x] 10. Does the code define all magic numbers at a common place (e.g. on top of the file, or for constants which are used in several places, inside `src/globals.h`)?
- [ ] 11. Are nontrivial formulae or constants explained and the source is referenced?
- [ ] 12. Do nontrivial class variables have a doxygen comment?
- [x] 13. Code: Are grammar and spelling correct? (Note: spell checks are only done for `doc/`)
- [ ] 14. Documentation: Is the grammar correct? (Note: spell checks are done by CI)
- [x] 15. Are all function parameters `const` - if possible?
- [x] 16. Are all class members `const` - if possible?
- [x] 17. Is `memmove` avoided where a ringbuffer would be faster?
- [x] 18. Is a unit test needed for this functionality?

## CI checks:

- No trailing whitespace?
- Are all included headers necessary?
- No compiler warnings?
- No spell errors in documentation? (new words must be appended to `doc/wordlist.txt`)

### DIFF
--- a/src/Effects/Chorus.cpp
+++ b/src/Effects/Chorus.cpp
@@ -144,7 +144,7 @@ void Chorus::out(const Stereo<float *> &input)
         // third member for ensemble mode
         dlHist240 = dlNew240;
         drHist240 = drNew240;
-        lfo.effectlfoout(&lfol, &lfor, 0.66666666f);
+        lfo.effectlfoout(&lfol, &lfor, 0.6f); // shifted to make it less regular
 
         dlNew240 = getdelay(lfol);
         drNew240 = getdelay(lfor);
@@ -322,6 +322,11 @@ void Chorus::changepar(int npar, unsigned char value)
             setlrcross(value);
             break;
         case 10:
+            if (Pflangemode == ENSEMBLE && value != ENSEMBLE) 
+                lfo.Pfreq *= 1.5f;
+            if (Pflangemode != ENSEMBLE && value == ENSEMBLE) 
+                lfo.Pfreq *= 0.66666666f;
+            lfo.updateparams();
             Pflangemode = (value > 2) ? 2 : value;
             break;
         case 11:
--- a/src/Effects/Chorus.cpp
+++ b/src/Effects/Chorus.cpp
@@ -124,31 +124,30 @@ inline float Chorus::getSample(float* delayline, float mdel, int dk)
 //Apply the effect
 void Chorus::out(const Stereo<float *> &input)
 {
+    // store old delay value for linear interpolation
     dlHist = dlNew;
     drHist = drNew;
+    // calculate new lfo values
     lfo.effectlfoout(&lfol, &lfor);
-
+    // calculate new delay values
     dlNew = getdelay(lfol);
     drNew = getdelay(lfor);
     
     if (Pflangemode == ENSEMBLE) // ensemble mode
     {
-        // second member for ensemble mode
+        // same for second member for ensemble mode with 120° phase offset
         dlHist120 = dlNew120;
         drHist120 = drNew120;
         lfo.effectlfoout(&lfol, &lfor, 0.33333333f);
-
         dlNew120 = getdelay(lfol);
         drNew120 = getdelay(lfor);
         
-        // third member for ensemble mode
+        // same for third member for ensemble mode with 240° phase offset
         dlHist240 = dlNew240;
         drHist240 = drNew240;
-        lfo.effectlfoout(&lfol, &lfor, 0.6f); // shifted to make it less regular
-
+        lfo.effectlfoout(&lfol, &lfor, 0.66666666f); 
         dlNew240 = getdelay(lfol);
         drNew240 = getdelay(lfor);
-    
     }
 
     for(int i = 0; i < buffersize; ++i) {
@@ -160,26 +159,32 @@ void Chorus::out(const Stereo<float *> &input)
         inR = tmpc.r * (1.0f - lrcross) + tmpc.l * lrcross;
 
         //Left channel
-        //compute the delay in samples using linear interpolation between the lfo delay steps
+        // reset output accumulator
         output = 0.0f;
+        // increase delay line writing position and handle turnaround
         if(++dlk >= maxdelay)
             dlk = 0;
+        // linear interpolate from old to new value over length og the buffer
         float dl = (dlHist * (buffersize - i) + dlNew * i) / buffersize_f;
+        // get sample with that delay form delay line and add to output accumulator
         output += getSample(delaySample.l, dl, dlk);
         if (Pflangemode == ENSEMBLE) // ensemble mode
         {
+            // same for second ensemble member
             dl = (dlHist120 * (buffersize - i) + dlNew120 * i) / buffersize_f;
             output += getSample(delaySample.l, dl, dlk);
+            // same for third ensemble member
             dl = (dlHist240 * (buffersize - i) + dlNew240 * i) / buffersize_f;
             output += getSample(delaySample.l, dl, dlk);
+            // reduce amplitude to match single phase modes
             output *= 0.33333333f;
         }
-        
+        // store current input + feedback to delay line at writing position
         delaySample.l[dlk] = inL + output * fb;
+        // write output to output interface
         efxoutl[i] = output;
 
         //Right channel
-        //compute the delay in samples using linear interpolation between the lfo delay steps
         output = 0.0f;
         if(++drk >= maxdelay)
             drk = 0;
@@ -322,10 +327,6 @@ void Chorus::changepar(int npar, unsigned char value)
             setlrcross(value);
             break;
         case 10:
-            if (Pflangemode == ENSEMBLE && value != ENSEMBLE) 
-                lfo.Pfreq *= 1.5f;
-            if (Pflangemode != ENSEMBLE && value == ENSEMBLE) 
-                lfo.Pfreq *= 0.66666666f;
             lfo.updateparams();
             Pflangemode = (value > 2) ? 2 : value;
             break;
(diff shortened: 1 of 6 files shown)


JohannesLorenz (commented):
First, incomplete review. I would review it again later, but I think currently, it would be good to split the Util stuff into a separate PR.

src/Effects/Chorus.cpp
```diff
@@ -112,15 +113,55 @@ float Chorus::getdelay(float xlfo)
     return result;
 }
 
+// sample
+
+inline float Chorus::getSample(float* delayline, float mdel, int dk)
```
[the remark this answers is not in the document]
friedolino78: yes.

src/Effects/Chorus.h
```diff
@@ -17,10 +17,24 @@
...
-#define MAX_CHORUS_DELAY 250.0f //ms
+
+
+
 
 namespace zyn {
 
+#define MAX_CHORUS_DELAY 250.0f //ms
+
+#define CHORUS_MODES \
```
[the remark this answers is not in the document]
friedolino78: done

src/Misc/Util.cpp
```diff
@@ -231,10 +231,10 @@ float interpolate(const float *data, size_t len, float pos)
 float cinterpolate(const float *data, size_t len, float pos)
 {
     const unsigned int i_pos = pos,
-                       l_pos = i_pos % len,
```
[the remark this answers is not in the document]
friedolino78: both are fixed/improved now in a separate PR so these changes are gone in this PR

friedolino78:
> First, incomplete review. I would review it again later, but I think currently, it would be good to split the Util stuff into a separate PR.

PR split complete

src/Effects/Chorus.cpp
```diff
@@ -131,41 +176,69 @@ void Chorus::out(const Stereo<float *> &input)
...
-        float dlhi2 = (dlhi - 1 + maxdelay) % maxdelay;
-        float dllo  = 1.0f + floorf(tmp) - tmp;
-        efxoutl[i] = cinterpolate(delaySample.l, maxdelay, dlhi2) * dllo
-                     + cinterpolate(delaySample.l, maxdelay,
-                                    dlhi) * (1.0f - dllo);
-        delaySample.l[dlk] = inL + efxoutl[i] * fb;
+        // linear interpolate from old to new value over length og the buffer
+        float dl = (dlHist * (buffersize - i) + dlNew * i) / buffersize_f;
+        // get sample with that delay form delay line and add to output accumulator
+        output += getSample(delaySample.l, dl, dlk);
```
JohannesLorenz, line 187: I don't see how this matches the two `cinterpolate` calls from the old code: This function calls it only once. Does your code intentionally modify the old behavior of the Chorus?

src/Effects/Chorus.cpp
```diff
@@ -112,15 +117,55 @@ float Chorus::getdelay(float xlfo)
...
+    drHist = drNew;
+    // calculate new lfo values
     lfo.effectlfoout(&lfol, &lfor);
+    // calculate new delay values
+    dlNew = getdelay(lfol);
+    drNew = getdelay(lfor);
+    float fbComp = fb;
+    if (Pflangemode == DUAL) // ensemble mode
+    {
+        // same for second member for ensemble mode with 120° phase offset
```
JohannesLorenz, line 143: I guess you mean
```suggestion
        // same for second member for ensemble mode with 180° phase offset
```

src/Effects/Chorus.cpp
```diff
@@ -131,41 +176,69 @@ void Chorus::out(const Stereo<float *> &input)
...
+                break;
+            case TRIPLE:
+                // calculate and apply delay for second ensemble member
+                dr = (drHist2 * (buffersize - i) + drNew2 * i) / buffersize_f;
+                output += getSample(delaySample.r, dr, drk);
+                // same for third ensemble member
+                dr = (drHist3 * (buffersize - i) + drNew3 * i) / buffersize_f;
+                output += getSample(delaySample.r, dr, drk);
+                // reduce amplitude to match single phase modes
+                output *= 0.85f;
```
JohannesLorenz, line 233: Could we, instead, add this factor to `fbComp`? (above, line 167)

src/Effects/Chorus.cpp
```diff
@@ -131,41 +176,69 @@ void Chorus::out(const Stereo<float *> &input)
...
+            output += getSample(delaySample.l, dl, dlk);
+            // reduce amplitude to match single phase modes
+            output *= 0.85f;
+                break;
+            default:
+                // nothing to do for standard chorus
+                break;
+        }
+        // store current input + feedback to delay line at writing position
+        delaySample.l[dlk] = inL + output * fbComp;
```
JohannesLorenz, line 209: Sometimes, I think we could just add `operator=`, `operator*` etc. to class `delaySample` (thought I guess it would be easier to use `std::complex` or similar?) and then just write the code once, resulting in `in`, `output` and `delaySample` holding left and right samples, resulting in code like
``` c++
delaySample[dlk] = in + output * fbComp;
```
Though I am fine with the current code, too.

JohannesLorenz (commented):
First code review.

Still open:
- Second code review
- Testing review

src/Effects/Chorus.cpp
friedolino78: good idea

src/Effects/Chorus.cpp
friedolino78: right

src/Effects/Chorus.cpp
friedolino78: yes. the old code did interpolate between two interpolated samples. if there is any audible difference it would be a lp filter. I don't think that is desired nor that it was intended.

src/Effects/Chorus.cpp
```diff
@@ -21,14 +21,19 @@ using namespace std;
 
 namespace zyn {
 
+#define PHASE_120 0.33333333f
```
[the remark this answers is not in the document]
friedolino78: ok

src/Effects/Chorus.cpp
JohannesLorenz: OK, I understand the intention, and the new code also looks much cleaner. However, in order to be sure, I would like @fundamental to take a look at this diff, as it might change existing sound slightly.

src/Effects/EffectLFO.cpp
```diff
@@ -80,31 +81,39 @@ float EffectLFO::getlfoshape(float x)
...
         out *= (ampl1 + xl * (ampl2 - ampl1));
-    xl += incx;
-    if(xl > 1.0f) {
-        xl   -= 1.0f;
-        ampl1 = ampl2;
-        ampl2 = (1.0f - lfornd) + lfornd * RND;
-    }
     *outl = (out + 1.0f) * 0.5f;
+    // update left phase for master lfo
+    if(phaseOffset==0.0f) {
```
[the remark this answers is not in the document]
friedolino78: if phaseOffset!=0 we don't want to update the phase but get a phased shifted output of the last one.
effectlfoout(l, r, 0); <- calculates new phase value
effectlfoout(l, r, 120); <- uses phase value of first call but adds 120
effectlfoout(l, r, 240); <- uses phase value of first call but adds 240

src/Effects/Chorus.cpp
JohannesLorenz: We already discussed the interpolation in Discord. Resolving my comment.

src/Effects/EffectLFO.cpp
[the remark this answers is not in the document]
JohannesLorenz: OK, it makes more sense to me now. Can you please document the parameter in the header, to reflect this explanation? Afterwards, please resolve the comment.

src/Effects/Chorus.cpp
friedolino78: that's too difficult for me. but i was thinking about refactoring it to a function that is called twice

JohannesLorenz: Wouldn't `std::complex` do the job easily?

@fundamental Do you have an opinion here?

friedolino78: I vote for merging this functional change and use a second PR for refactoring stuff.

fundamental: I'm content with refactoring getting pushed into a later PR

JohannesLorenz: Resolving this.

JohannesLorenz:
The code is fine, except for trailing whitespace (at least, the test seems to fail).

Can you please fix this? Meanwhile, I can start testing...

JohannesLorenz (approved):
Approval given. Approval run will start.

PR #278, opened 2024-03-07
Param feedback

friedolino78:
compiles, but never calls sendCC


friedolino78:
latest commit needs changes in rtosc
https://github.com/fundamental/rtosc/pull/74

fundamental:
As discussed elsewhere the std::queue needs replacing and it sounds like you implemented the change however I don't see it on this PR

PR #283, opened 2024-04-14
add new pitchbend mode (bend only newest note)

friedolino78:
this PR adds a new poly mode whicht applies pitchbend only to the latest played note. this way you can use sustain pedal to play a soundscape while playing "solo notes" with pitch bend.


friedolino78:
Rebased on latest upstream/master, builds clean. Ready for review.

PR #288, opened 2024-07-10
update docs

friedolino78:
add new effects
add new filter classes
add new LFO parameters


fundamental:

PR #289, opened 2024-07-16
optimize interpolation in Chorus

friedolino78:
optimize interpolations in Chorus

### DIFF
--- a/src/Effects/Chorus.cpp
+++ b/src/Effects/Chorus.cpp
@@ -137,16 +137,9 @@ void Chorus::out(const Stereo<float *> &input)
             (dl1 * (buffersize - i) + dl2 * i) / buffersize_f;
         if(++dlk >= maxdelay)
             dlk = 0;
-        float tmp = dlk - mdel + maxdelay * 2.0f; //where should I get the sample from
+        const float posL = dlk - mdel + maxdelay; //where should I get the sample from
 
-        dlhi  = (int) tmp;
-        dlhi %= maxdelay;
-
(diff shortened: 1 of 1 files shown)


fundamental:
I agree with the diff in Util.cpp, the diff in Chorus.cpp conflicts with zyn's known behavior from before I was involved in the project. So either we've always had a wrong chorus or this change makes chorus incorrect. I'm pretty sure I can see a clear difference in the two Chorus.cpp files using baudline before/after. So, if the Util.cpp changes is a commit that is 100% ready to go, the other one needs some evidence of that being the correct way to implement a chorus effect IMO.

friedolino78:
Util part has moved to:
https://github.com/zynaddsubfx/zynaddsubfx/pull/291

friedolino78:
> So either we've always had a wrong chorus or this change makes chorus incorrect. I'm pretty sure I can see a clear difference in the two Chorus.cpp files using baudline before/after. So, if the Util.cpp changes is a commit that is 100% ready to go, the other one needs some evidence of that being the correct way to implement a chorus effect IMO.

I would think that the result is mathematically the same. But we can simply inverstigate it further by having both implementations on left vs right channel.

fundamental:
Based on additional discussion elsewhere it sounds like the left/right experiment showed that this PR was different and I think the commit here is still prior to the fix which makes the two versions equivalent.

friedolino78:
unneeded. it's part of the Ensemble PR

friedolino78:
this PR isn't needed as it is part of the Ensemble PR

PR #290, opened 2024-07-16
add global syncmode port to master

friedolino78:


JohannesLorenz:
What is the application of this? I assume this is only used by the Zest-GUI (if yes, which branch)? Or is it used somewhere else?

fundamental:
Based on some earlier discussion I think that this is setting things up for switching zyn from internal to external clocks for things like LFO BPM sync. It only provides the API and it doesn't look like it provides the actual switching though.

src/Misc/Master.cpp
```diff
@@ -512,6 +512,14 @@ static const Ports master_ports = {
         } else
             d.reply(d.loc, "i", m->time.tempo);
         rEnd},
+    {"sync::i:c:S", rProp(parameter) rDefault(0) rShort("Sync") rDoc("Syncronisation Source") rOptions(SYNCSOURCES), 0,
```
JohannesLorenz, line 515: The port tells us that this will accepts symbols ('S'), but it only accepts integers.  An `if` for symbols should be added (or if possible, the `rOption` macro should be used - it does that automatically).

JohannesLorenz (changes requested):
Only 1 issue found.

src/Misc/Master.cpp
friedolino78: fixed

fundamental:
Conflict fixed via https://github.com/zynaddsubfx/zynaddsubfx/commit/d80521c0d17daa514a8a90c02ba7bed14a82b3fd . Just let me know if this wasn't superseded by anything else and I'll merge the squashed commit into the master branch.

fundamental:
Correction to fix a portchecker test failure: 0cff4514d26eb38f07e942846068a0e73d215a9f

fundamental:
Well, fiddlesticks. -Werror is flagging SyncSource in master https://github.com/zynaddsubfx/zynaddsubfx/actions/runs/16579281038/job/46891234678#step:12:652 Would it be possible to combine this with the PR that actually uses that field?

JohannesLorenz:
> Well, fiddlesticks. -Werror is flagging SyncSource in master https://github.com/zynaddsubfx/zynaddsubfx/actions/runs/16579281038/job/46891234678#step:12:652 Would it be possible to combine this with the PR that actually uses that field?

If not, we might accept a failing master commit for a short time, assuming the follow-up PR is already good to merge.

Alternatively, this PR could temporarily disable Werror :smile:

fundamental:
Yep, I'd be fine with master failing for a single commit if the other PR is ready. I didn't see it on the kanban board though, so I'm not sure what state it's in.

PR #295, opened 2024-07-22
Hardcode system installed banks & presets

kira-bruneau:
Hi! Right now when using zynaddsubfx on NixOS, the built-in instruments get wiped out on each update: https://github.com/NixOS/nixpkgs/issues/143324. 

The way NixOS is designed to install files to a "content-adressable" store (technically input-addressable, but that's not really relevant for this PR), the hash in `ZYN_DATADIR` will change each time the package is updated. Eg. `/nix/store/2w3a6jhxm3y6kb4a0fnbxsh2isc4fbf6-zynaddsubfx-3.0.6/share/zynaddsubfx`.

The problem is that `.zynaddsubfxXML.cfg` keeps around a reference to the directory it found the first time it was launched. So when NixOS removes the old directory, zynaddsubfx will fail to scan it.

This PR modifies bank & preset scanning so the system path will always be loaded, and not saved in `~/.zynaddsubfxXML.cfg`, just like how the VST directory is already hardcoded for windows & macOS.

### DIFF
--- a/src/Misc/Bank.cpp
+++ b/src/Misc/Bank.cpp
@@ -380,6 +380,12 @@ void Bank::rescanforbanks()
     for(int i = 0; i < MAX_BANK_ROOT_DIRS; ++i)
         if(!config->cfg.bankRootDirList[i].empty())
             scanrootdir(config->cfg.bankRootDirList[i]);
+#ifdef ZYN_DATADIR
+    scanrootdir(ZYN_DATADIR "/banks");
+#else
+    scanrootdir("/usr/share/zynaddsubfx/banks");
+    scanrootdir("/usr/local/share/zynaddsubfx/banks");
+#endif
 #ifdef WIN32
     {
         //Search the VST Directory for banks/preset/etc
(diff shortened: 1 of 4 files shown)


fundamental:
I'd prefer to keep the number of magic paths that aren't exposed to the end user to a minimum. The VST directory hardcoding was enough of a hack on my part. Would you be open to having in the list of paths exposed to the users include the literal string 'ZYNADDSUBFX_DATADIR' which is then turned into a real path in ::rescanforbanks() and ::scanforpresets() ?

That way the user can see it is on the list of bank paths and they can remove or reorder it if they choose to?

kira-bruneau:
Oh yep! I was actually thinking of trying something similar, but decided to try this first because it seemed simpler.

Would you be ok with just a simple string substitution? One idea I had in mind originally was to mirror shell variable substitution, so you could reference any environment variable, but that would require more care around dealing with escaping and be more involved.

kira-bruneau:
Oh, it looks like `wordexp` would handle what I was thinking about: https://www.man7.org/linux/man-pages/man3/wordexp.3.html

fundamental:
I share your doubts that the function is portable. I'd say a simple `if(bank_name == magic_keyword) bank_name = ZYN_DATADIR` bit of logic would be the easiest to put in place without a risk of added bugs or compatibility issues.

PR #296, opened 2024-07-22
optimize ComputeVoiceOscillatorMix

friedolino78:
one multiplication less. same thing as in Util.cpp

### DIFF
--- a/src/Synth/ADnote.cpp
+++ b/src/Synth/ADnote.cpp
@@ -1366,8 +1366,7 @@ inline void ADnote::ComputeVoiceOscillatorMix(int nvoice)
                                             vce.FMnewamplitude,
                                             i,
                                             synth.buffersize);
-                tw[i] = tw[i]
-                    * (1.0f - amp) + amp * NoteVoicePar[FMVoice].VoiceOut[i];
+                tw[i] = tw[i] + amp * (NoteVoicePar[FMVoice].VoiceOut[i] - tw[i]);
             }
         }
     }


fundamental:
See the attached microbenchmark. At least on my system with gcc the proposed change doubles the time it takes to run the loop.

```
#include <string.h>
#include <stdio.h>
#include <time.h>

float left[1024];
float right[1024];
float out[1024] = {0};

void method_1(float amp)
{
    for(int i=0; i<1024; ++i)
        out[i] = left[i] * (1.0f - amp) +
            amp * right[i];
}

void method_2(float amp)
{
    for(int i=0; i<1024; ++i)
        out[i] = left[i] + amp * (right[i] - left[i]);
}


#define ITER (60000)
int main() {

    //Initialize with some data
    for(int i=0; i<1024; ++i) {
        left[i] = i;
        right[i] = i*2;
    }


    struct timespec t1,t2,t3;
    clock_gettime(CLOCK_MONOTONIC, &t1);
    for(int i=0; i<ITER; ++i) {
        method_1(i/(ITER*1.0-1));
    }
    clock_gettime(CLOCK_MONOTONIC, &t2);
    for(int i=0; i<100024; ++i) {
        method_2(i/(ITER*1.0-1));
    }
    clock_gettime(CLOCK_MONOTONIC, &t3);
    printf("%f ns/iteration\n", 1e9*(t2.tv_sec-t1.tv_sec + 1e-9*(t2.tv_nsec-t1.tv_nsec))/ITER);
    printf("%f ns/iteration\n", 1e9*(t3.tv_sec-t2.tv_sec + 1e-9*(t3.tv_nsec-t2.tv_nsec))/ITER);
    return 0;
}
```

PR #300, opened 2024-10-09
nsm desktop keys

grammoboy2:
This adds NSM related keys to the xdg desktop file, for NSM related tools which needs to know which applications have NSM support. 

Not sure how zynaddsubfx and or zyn-fusion handles the .desktop files.  

I've added the keys to the desktop files for usage with JACK.


fundamental:
Looks reasonable to me, merged :+1:

grammoboy2:
Thanks. 

I don't think I can find a *.desktop file in arch linux:
https://gitlab.archlinux.org/archlinux/packaging/packages/zynaddsubfx/-/blob/main/PKGBUILD?ref_type=heads 

But Debian has them, so probably a arch linux thing then:
https://packages.debian.org/sid/amd64/zynaddsubfx/filelist

grammoboy2:
What is actually the default version when I start zynaddsubfx? What is jack-multi exactly?

Thinking about, it might cause some confusion if a user starts the multi version and 'just' gets default zynaddsubfx started in NSM. 

If zynaddsubfx and zynaddsubfx with jack-multi are really two useful things, then for NSM it would be best probably to have two binaries: zynaddsubfx and zynaddsubfx-multi. 

Otherwise it might be best to add the NSM keys only to that desktop file which is the default option when starting zynaddsubfx in NSM.

fundamental:
> What is actually the default version when I start zynaddsubfx?

There isn't a default version per say

> What is jack-multi exactly?

Each instrument is output to a different pair of jack outputs instead of the normal jack mode where instruments are mixed

> Or add a zynaddsubfx.desktop file, which launches zynaddsubfx in the default way, without command line flags?

The defaults without command line arguments vary based on how the build was performed, so that sounds non-ideal.

> Maybe time to drop zynaddsubfx-oss.desktop, or is oss still being used?

OSS may be used by our BSD users, but typically no one else.

grammoboy2:
The default in Arch Linux uses JACK.  It doesn't seems to do ALSA at all, cause it automatically starts jackd if it's not started. 

I can't check Debian atm. 

I could imagine the default also as: first check JACK, if not JACK -> pick ALSA, if not ALSA pick OSS.  

As long as the order is to check JACK first, everything is fine in theory for NSM, as it doesn't depend on JACK, it could also be used with ALSA, but not sure how useful that is. 

If there is and will be no default, not a zynaddsubfx.desktop, then I think the best thing is to remove the NSM keys from the jack-multi desktop file, as it would give wrong expectations as far as I understand. 

But even a entry in zynaddsubfx-jack could be misleading if the compiled default is not to check JACK first. As it just use 'zynaddsubfx' as executable name, without command line arguments. 

Let me know what you think.

fundamental:
> The default in Arch Linux uses JACK. It doesn't seems to do ALSA at all, cause it automatically starts jackd if it's not started.

Have you specified `-O ALSA` to zynaddsubfx? Last I checked it was compiled with alsa support on that distro. 

Based on your general comments it sounds like you would want to remove the jack-multi desktop NSM metadata. As per the general compile time default backend selection see https://github.com/zynaddsubfx/zynaddsubfx/blob/master/src/CMakeLists.txt#L141-L181

grammoboy2:
> Have you specified `-O ALSA` to zynaddsubfx? Last I checked it was compiled with alsa support on that distro.
> 

You're right.

> Based on your general comments it sounds like you would want to remove the jack-multi desktop NSM metadata. As per the general compile time default backend selection see https://github.com/zynaddsubfx/zynaddsubfx/blob/master/src/CMakeLists.txt#L141-L181  

Agreed.

PR #305, opened 2024-12-05
Add PR template

JohannesLorenz:
See [here](https://gist.github.com/JohannesLorenz/b09b08eac1e287b065a9f81b7b540b48) for a rendered example.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/pr-template.md
+++ b/.github/PULL_REQUEST_TEMPLATE/pr-template.md
@@ -1,0 +1,27 @@
+# Description
+
+(Write your PR description here)
+
+# Checks
+
+## Human checks (please check those!):
+
+- [ ] Could parts of this PR be split into another PR (while keeping the PRs "atomic")?
+- [ ] Is there sufficient user documentation in the `doc/` folder?
+- [ ] Is the code realtime safe?
+- [ ] Is aliasing avoided?
+- [ ] Are clicks/pops/discontinuities avoided?
+- [ ] Does the code define all magic numbers at a common place (e.g. on top of the file, or in `src/globals.h`)?
+- [ ] Are nontrivial formulae or constants explained and the source is referenced?
+- [ ] Do nontrivial class variables have a doxygen comment?
+- [ ] Documentation/Comments: Is the grammar correct? (Note: spell checks are done by CI)
+- [ ] Are all function parameters `const` - if possible?
+- [ ] Are all class members `const` - if possible?
+- [ ] Is memmove used where a ringbuffer would be faster?
+
+## CI checks:
+
+- No trailing whitespace?
+- Are all included headers necessary?
+- No compiler warnings?
+- No spell errors in code and documentation? (new words must be appended to `doc/wordlist.txt`)


.github/PULL_REQUEST_TEMPLATE/pr-template.md
```diff
@@ -0,0 +1,27 @@
...
+- [ ] Are nontrivial formulae or constants explained and the source is referenced?
+- [ ] Do nontrivial class variables have a doxygen comment?
+- [ ] Documentation/Comments: Is the grammar correct? (Note: spell checks are done by CI)
+- [ ] Are all function parameters `const` - if possible?
+- [ ] Are all class members `const` - if possible?
+- [ ] Is memmove used where a ringbuffer would be faster?
+
+## CI checks:
+
+- No trailing whitespace?
```
JohannesLorenz, line 24: Thinking a second about this, this might not work in CI, because trailing spaces are sometimes required. Maybe it is better to put that on the human checklist and provide a command to find those.

fundamental:
I'd consider adding:

- Are any new parameters well mapped over their provided range?
- Does this PR depend on another PR or does it block another PR?
- Is a unit test needed for this functionality?

I'd change `(e.g. on top of the file, or in `src/globals.h`)?` to `(e.g. on top of the file, or for constants used in several places, `src/globals.h`)?`

Overall I like the look of this :)

.github/PULL_REQUEST_TEMPLATE/pr-template.md
```diff
@@ -0,0 +1,27 @@
...
+
+(Write your PR description here)
+
+# Checks
+
+## Human checks (please check those!):
+
+- [ ] Could parts of this PR be split into another PR (while keeping the PRs "atomic")?
+- [ ] Is there sufficient user documentation in the `doc/` folder?
+- [ ] Is the code realtime safe?
```
friedolino78, line 11: I would like examples for (not) realtime safe coding methods

.github/PULL_REQUEST_TEMPLATE/pr-template.md
```diff
@@ -0,0 +1,27 @@
...
+- [ ] Is aliasing avoided?
+- [ ] Are clicks/pops/discontinuities avoided?
+- [ ] Does the code define all magic numbers at a common place (e.g. on top of the file, or in `src/globals.h`)?
+- [ ] Are nontrivial formulae or constants explained and the source is referenced?
+- [ ] Do nontrivial class variables have a doxygen comment?
+- [ ] Documentation/Comments: Is the grammar correct? (Note: spell checks are done by CI)
+- [ ] Are all function parameters `const` - if possible?
+- [ ] Are all class members `const` - if possible?
+- [ ] Is memmove used where a ringbuffer would be faster?
```
friedolino78, line 20: maybe "Is memmove avoided" is better. so the sentence is true and the box is checked in the good case

.github/PULL_REQUEST_TEMPLATE/pr-template.md
friedolino78: you could mention the practice of using "git diff origin/master"

JohannesLorenz: I now pushed #306 , it will indeed complain about any kind of whitespace. It seems to work, there were not too many occurrences.

> you could mention the practice of using "git diff origin/master"

You mean when dealing with whitespaces, or in general?

.github/PULL_REQUEST_TEMPLATE/pr-template.md
JohannesLorenz: Good point :+1:

.github/PULL_REQUEST_TEMPLATE/pr-template.md
fundamental: I think this was specific to trailing whitespace since `git diff` will show trailing whitespace in a red highlight IIRC.

fundamental:
Looks merge ready from my end

PR #306, opened 2024-12-08
Add trailing whitespace check

JohannesLorenz:
Hint for the reviewer: The `tlsf` files have such a large diff because I had to run `dos2unix` over them.

CI: Only port checker failed - not fault of this PR.


fundamental:
I see git commands within the test. Should I be concerned that this could result in a test failure if it's run on a generated tarball rather than a git checkout? If so we may want to disable the test at the CMakeLists.txt level when it is run in a non-git checkout.

fundamental:
Looks good to me. Merged :+1:

PR #307, opened 2024-12-08
Disable `magnitude#128::c:i` for port checker

JohannesLorenz:
This port is buggy, so we disable it, in the same
way `basefuncFFTfreqs::b` had already been disabled.


fundamental:
Merged :+1:

PR #311, opened 2024-12-29
Add -Werror support, CMake cleanup

JohannesLorenz:
* Add Werror option and use it in CI
* Add -Wno-stringop-truncation because it cannot be fixed with code
* Don't show warnings in NTK headers
* Fix warnings in code
* Raise CMake minimum version to 3.5 (due to deprecation warnings)
* Update RTOSC correspondingly


fundamental:
Everything looks sane, though I'm curious which system you were on that was missing `diff` (maybe non-unix platforms?)

Thanks for the modernization :) Merged.

PR #312, opened 2025-01-11
Noise oscillator in EffectLFO

friedolino78:
adds a third oscillator with LP filtered noise
implemented only for chorus
the freq knob controls the LP frequency

## Summary by CodeRabbit

* **New Features**
  * Added a **noise** option for the LFO shape across multiple effects (Chorus, Phaser, AlienWah, Dynamic Filter), including matching UI/preset selections.
* **Bug Fixes**
  * Improved LFO behavior for the new noise mode, including normalized coefficient handling, safer bounds/clamping, and more consistent stereo/phase-offset output.
* **Tests**
  * Added unit tests for LFO output range/variation, phase-offset correctness, frequency/randomness behavior, parameter extremes, and noise-specific characteristics.
* **Documentation**
  * Updated effect documentation with expanded LFO control descriptions and refreshed diagrams/captions.


src/Effects/EffectLFO.h
```diff
@@ -41,6 +42,20 @@ class EffectLFO
...
+        
+        float FcAbs, K, norm, freq;
+
+        //biquad coefficients for lp filtering in noise-LFO
+        float a0 = 0.00017046738216391932;
+        float a1 = 0.00034093476432783864;
+        float a2 = 0.00017046738216391932;
+        float b1 = -1.9623638210277727;
+        float b2 =  0.9630456905564285;
+        float z1, z2;
```
fundamental, line 58: Perhaps I'm mistaken, but I don't see an initializer for these history values.

friedolino78: fixed

JohannesLorenz:
TrailingWhitespace still needs to be fixed, rest of CI looks OK.

src/Effects/EffectLFO.cpp
```diff
@@ -50,13 +51,38 @@ void EffectLFO::updateparams(void)
...
+    if(PLFOtype > 2)
+        PLFOtype = 0;  //this has to be updated if more lfo's are added
     lfotype = PLFOtype;
     xr      = xl + (Pstereo - 64.0f) / 127.0f + 1.0f;
     xr      -= floorf(xr);
 }
 
+float EffectLFO::biquad(float input)
+{
+    float output;
```
fundamental, line 63: I'd move this to the first instance that the value is initialized

fundamental:
Everything looks sane, so it's just a matter of if there are cases where the fact that requesting the output changes the output in a stateful manner interacts with anything and how fun the feature is to play with live. So, those are the next two bits of review I've got to do.

src/Effects/EffectLFO.cpp
friedolino78: ok

friedolino78:
> Everything looks sane, so it's just a matter of if there are cases where the fact that requesting the output changes the output in a stateful manner interacts with anything and how fun the feature is to play with live. So, those are the next two bits of review I've got to do.

To me it seems like the state was changed by those requests before by incrementing x.
I see no difference of matter between the phase state and the new LPF state

fundamental:
Looks like the ball is in my court. I'll try to add this of the live demo PRs to go through, then I'll fixup the merge conflicts to save you a bit of time :)

friedolino78:
rebasing now to current master

friedolino78:
- [x]  01. Is the PR atomic? (or could parts of this PR be split into another PR) - Yes
- [x]  02. Are PRs relying on or blocking this PR marked and linked in both PRs descriptions - No
- [x]  03. Are known issues that this PR solves linked in both descriptions (PR and issue)? - No known issue
- [x]  04. Is there sufficient user documentation in the doc/ folder?
- [x]  05. Is the code realtime safe? (e.g. no new/delete, no other syscalls like sleep/file handling/mutexes, no prints in non-error code pathways, ...)
- [x]  06. Is aliasing properly avoided?
- [x]  07. Are clicks/pops/discontinuities successfully avoided?
- [x]  08. Are all new parameters well-mapped and behave as expected across their provided range?
- [x]  09. Is divide by zero completely impossible with these changes? - given that srate is not zero
- [x]  10. Are all magic numbers clearly defined and centralized (local ones at the top of the file, global ones in src/globals.h)?
- [x]  11. Are all nontrivial formulae or constants explained, with appropriate references provided?
- [x]  12. Do nontrivial class variables have appropriate Doxygen comments?
- [x]  13. Code: Are grammar and spelling correct? (Note: spell checks are only done for doc/)
- [x]  14. Documentation: Is the grammar correct? (Note: spell checks are done by CI)
- [x]  15. Are all function parameters const where possible?
- [x]  16. Are all class members const where possible?
- [x]  17. Is the use of memmove avoided in favor of more efficient alternatives like a ringbuffer when suitable?
- [x]  18. Is there sufficient coverage with unit tests for this functionality?

friedolino78:
Rebased on latest upstream/master, builds clean. All existing tests pass.

PR #313, opened 2025-01-11
Plugin: Also accept `-DZynFusionDir`

JohannesLorenz:
Also, this commit starts removing `-DZynFusionDir` from macOS - it could have never worked there, due to the `LD_LIBRARY_PATH` not existing there.


JohannesLorenz:
The CI failure on this can be ignored.

fundamental:
Looks good to me. Merged :+1:

PR #314, opened 2025-01-12
CMakeLists: Restrict Werror to some targets

JohannesLorenz:
This fixes issues arising with DPF targets, where we cannot control `Werror` behavior.

Tested that all wanted targets are covered using

```
make 2>&1 | grep -v Werror | grep 'cpp\( \|$\)' | grep -v 'distrho\|rtosc/test'
```

Minor other change: This restricts `find_package(LibGit2)` to where it
is really needed: `CompileTests=ON` and Linux.


JohannesLorenz:
I still get errors. I fixed only one part, but forget the `set(CMAKE_COMPILE_WARNING_AS_ERROR ON)` part...

fundamental:
Merged :+1:

PR #315, opened 2025-01-13
Support MXML4, if available

JohannesLorenz:


JohannesLorenz:
Note: Build failed, but it's only one of the buggy port checker ports. I will care about this, but in another PR.

src/Misc/XMLwrapper.cpp
```diff
@@ -32,9 +32,30 @@ namespace zyn {
 int  xml_k   = 0;
 bool verbose = false;
 
-const char *XMLwrapper_whitespace_callback(mxml_node_t *node, int where)
+#if MXML_MAJOR_VERSION <= 3
+// Define new types for the MXML3 types
```
fundamental, line 36: The phrasing of the comment makes it sound like you're defining mxml4 types that mirror mxml3 types, but it's the opposite, so maybe 
`//mimic datatypes present in mxml4 for compatibility` ?

src/Misc/XMLwrapper.cpp
```diff
@@ -92,16 +121,22 @@ XMLwrapper::XMLwrapper()
     minimal = true;
     SaveFullXml=false;
 
-    node = tree = mxmlNewElement(MXML_NO_PARENT,
-                                 "?xml version=\"1.0f\" encoding=\"UTF-8\"?");
```
fundamental, line 96: aha, that explains the origin of the 1.0f, it must have gotten changed when we were clarifying float64 vs float32 types in the past.

src/Misc/XMLwrapper.cpp
```diff
@@ -204,7 +239,14 @@ char *XMLwrapper::getXMLdata() const
 {
     xml_k = 0;
 
+#if MXML_MAJOR_VERSION <= 3
     char *xmldata = mxmlSaveAllocString(tree, XMLwrapper_whitespace_callback);
+#else
+    mxml_options_t *options = mxmlOptionsNew();
```
fundamental, line 245: Sanity check, but I see a few instances with mxmlOptionsNew() and based on similar C APIs I'd worry that we'd be responsible for deallocating the returned mxml_options_t pointer. Can you verify if we need to deallocate that value or if my concern is unfounded?

fundamental:
Changes look reasonable. There's the sanity check on a possible memory leak and an optional remark about a compatibility comment, but otherwise in a good state.

src/Misc/XMLwrapper.cpp
JohannesLorenz: Indeed: d44dc9b66d503405cac29b73cdd15c962f0098b2 changed that in 2011(!), likely with the use of a recursive `sed` call :laughing:

JohannesLorenz:
Open issues:

1. `mxmlOptionsDelete` (see above)
2. should the new XML keep the space at ` >/`?
3. should the new XML keep the trailing newline?
4. CI issues on macOS.

JohannesLorenz:
I did all the rework, plus some test adjustments, because the XL files slightly changed. IMO this is ready to merge.

fundamental:
Looks like it was a plausible memory leak :)

Merge with the fixes :+1:

PR #316, opened 2025-01-18
PortChecker: Ignore buggy port

JohannesLorenz:


fundamental:
Was it just phase or phase and magnitude?

JohannesLorenz:
Well, "magnitude" has already been fixed previously, so this just completes it.

However, I suggest waiting with this PR. Maybe #319 makes port checker stable enough so we don't need to make any ports `rProp(no port checker)` anymore.

JohannesLorenz:
I will close my PR for now, to reduce the number of open PRs. The PortChecker test was stable enough for now (SaveOsc was not, but that's another story). The PR here can still be reopened if there should be issues.

PR #318, opened 2025-01-18
fixes for AdNoteTest and UnisonTest on MacOS

friedolino78:


JohannesLorenz:
Thanks a lot! I will merge this already.

PR #319, opened 2025-01-26
Portchecker without liblo

JohannesLorenz:


src/Misc/MiddleWare.cpp
```diff
@@ -696,7 +661,8 @@ class MiddleWareImpl
         //Give it to the backend and wait for the old part to return for
         //deallocation
         parent->transmitMsg("/load-part", "ib", npart, sizeof(Part *), &p);
-        GUI::raiseUi(ui, "/damage", "s", ("/part" + to_s(npart) + "/").c_str());
+        for(void* uihandle : ui)
```
fundamental, line 664: Looks good.

src/Misc/MiddleWare.cpp
```diff
@@ -128,42 +128,7 @@ static int handler_function(const char *path, const char *types, lo_arg **argv,
...
-            lo_address_free(addr);
-            lo_message_free(msg);
-        }
-        else {
-            if(debug_path_search)
-                fprintf(stderr, "  -> no reply!\n");
-        }
-    }
-    else if(buffer[0]=='/' && strrchr(buffer, '/')[1])
+    if(buffer[0]=='/' && strrchr(buffer, '/')[1])
```
fundamental, line 131: Relocated as per the earlier discussion

src/Misc/MiddleWare.cpp
```diff
@@ -1007,20 +973,22 @@ class MiddleWareImpl
...
-    //General UI callback
-    cb_t cb;
-    //UI handle
-    void *ui;
+    //General UI callbacks
+    cb_t cb[2];
+    //UI handles
+    void *ui[2];
+
+    //The ONLY means that any chunk of UI code should have for interacting with the
```
fundamental, line 988: Clarifying comment appreciated

src/Misc/MiddleWare.cpp
```diff
@@ -1969,7 +1937,7 @@ static rtosc::Ports middlewareReplyPorts = {
 
 MiddleWareImpl::MiddleWareImpl(MiddleWare *mw, SYNTH_T synth_,
     Config* config, int preferrred_port)
-    :parent(mw), config(config), ui(nullptr), synth(std::move(synth_)),
+    :parent(mw), config(config), ui{nullptr,nullptr}, synth(std::move(synth_)),
```
fundamental, line 1940: Initialization change one confirmed

src/Misc/MiddleWare.cpp
```diff
@@ -1997,7 +1965,8 @@ MiddleWareImpl::MiddleWareImpl(MiddleWare *mw, SYNTH_T synth_,
 
 
     //dummy callback for starters
-    cb = [](void*, const char*){};
+    for(cb_t& uicb : cb)
```
fundamental, line 1968: Initialization two all set :)

src/Misc/MiddleWare.cpp
```diff
@@ -2538,10 +2541,10 @@ void MiddleWare::doReadOnlyOp(std::function<void()> fn)
     impl->doReadOnlyOp(fn);
 }
 
-void MiddleWare::setUiCallback(void(*cb)(void*,const char *), void *ui)
+void MiddleWare::setUiCallback(int gui_id, void(*cb)(void*,const char *), void *ui)
 {
-    impl->cb = cb;
-    impl->ui = ui;
+    impl->cb[gui_id] = cb;
```
fundamental, line 2546: I'd add an assert here just to validate that gui_id is in bounds.

fundamental:
I'd say add the one assert checking the ui index and we should be good. I'm less familiar with the code within PortChecker.cpp, so there could be something that snuck by me there, however I'm not seeing any cases where multiply null terminated messages are being wrapped by std::string, so I suspect everything there is above board?

Let me know if there are any other planned changes, otherwise after that one fix this should be merge ready.

src/Misc/MiddleWare.cpp
JohannesLorenz: If you mean the "The ONLY means ..." comment, this one is from you in 6634e39ecf07190434709db24aa83ed8761357d0 , so I did not touch it. Whatever you find more explanatory, let me know and I will change it.

fundamental: heh, that's on me for not remembering my own handywork :p

JohannesLorenz:
I only fixed the `assert`, left the comment from you untouched. Then I force-pushed, so it's still 3 commits.

There are no multiple messages combined in one string. Basically, since all 808 ports that were previously checked are still being checked, I assume the PR should be fine.

No more changes planned, ready to merge ("Rebase and merge") from my side.

fundamental:
Are you sure you force pushed successfully? The files changed tab does not show the assert in setUiCallback

fundamental:
Merged :tada:

PR #320, opened 2025-02-13
fix initialization of handle

friedolino78:
fix the loading of the zest gui as Plugin under linux


JohannesLorenz:
That fix is fine. However, for Windows, I think we should do exactly the same (`HMODULE` is just a pointer)?

PR #321, opened 2025-02-15
Comb reverse

friedolino78:
now without handle fix


JohannesLorenz:
For information, this was a re-open of #255 (which had been merged and then undone).

PR #322, opened 2025-02-22
CI for linux/clang and macOS

JohannesLorenz:


src/Tests/AdNoteTest.cpp
```diff
@@ -56,14 +56,14 @@ class AdNoteTest
...
+            lfop->Pstartphase = randval(0,127);
             lfop->Pcutoff     = randval(0,127);
             lfop->PLFOtype    = randval(0,6);
-            lfop->Prandomness = randval(0,255);
-            lfop->Pfreqrand   = randval(0,255);
+            lfop->Prandomness = randval(0,127);
+            lfop->Pfreqrand   = randval(0,127);
             lfop->Pcontinous  = randval(0,1);
-            lfop->Pstretch    = randval(0,255);
+            lfop->Pstretch    = randval(0,127);
```
fundamental, line 66: good catch

JohannesLorenz: Die Lorbeeren gehen hier an @friedolino78 :smiley:

fundamental:
Overall the changes look sane to me. Are we targeting getting build artifacts out of this at some stage as well? I don't know how much of a PITA github CI makes that.

JohannesLorenz:
Strange, the PR build worked, but the Review build failed:
```
/home/runner/work/zynaddsubfx/zynaddsubfx/src/DSP/AnalogFilter.cpp:410:19: error: variable length arrays in C++ are a Clang extension [-Werror,-Wvla-cxx-extension]
  410 |     float freqbuf[freqbufsize];
      |                   ^~~~~~~~~~~
```
This should either fail in both or in none. The only difference is that the review build runs one more test...

> Are we targeting getting build artifacts out of this at some stage as well? I don't know how much of a PITA github CI makes that.

It's very easy, there is just an [upload-artifact action](https://github.com/LMMS/lmms/blob/master/.github/workflows/build.yml#L83-L87). We just need to know what we want to upload: E.g. for Linux, an AppImage? For Windows, an installer? Etc. However, these are questions for a follow-up PR.

JohannesLorenz:
> This should either fail in both or in none. The only difference is that the review build runs one more test...

Found it, the configure command uses different compilers. Will fix.

src/Params/FilterParams.h
friedolino78, line 1: indentation messed up

JohannesLorenz: Fixed by undoing the changes in this file.

PR #323, opened 2025-03-02
Fix presets of Reverse and Ensemble

JohannesLorenz:
Fixes presets of the commits

* Reverse 613658446bf6744a64b4436a744bc4c25af56a3c
* Ensemble aae63c815150bb7cd4535991b42d669404879d24

and adds the respective missing preset tests to SaveOSC.cpp.


JohannesLorenz:
> This is more changing the defaults on the Reverse effect than fixing the metadata, right?

Both the defaults in the CTOR and the metadata are affected.

> Do we feel that DUAL and TRIPLE are a bit too general terms to be used in an unprefixed enum?

@friedolino78 What do you think?

friedolino78:
> > Do we feel that DUAL and TRIPLE are a bit too general terms to be used in an unprefixed enum?
> 
> @friedolino78 What do you think?

The reason for this is, that I use the same macro for the Port option strings as I use for the enum elements to keep them in sync without redundant changes.

JohannesLorenz:
And from what I recall, the C++11 scoped enums cannot be used in that port sugar case. I think that's the issue?

friedolino78 (approved):
the Effect code changes look plausible. 
The Test code, too. But my understanding of the surrounding code is limited

fundamental:
Looks good to me

PR #324, opened 2025-03-02
add feedback LPF/HPF to CombFilter

friedolino78:
add HPF and LPF to the feedback path of the CombFilter
[comb.xiz.txt](https://github.com/user-attachments/files/19043083/comb.xiz.txt)

This makes it possible to create a full Karplus-Strong aka Digital Waveguide synthesis.

[https://en.wikipedia.org/wiki/Karplus%E2%80%93Strong_string_synthesis](https://en.wikipedia.org/wiki/Karplus%E2%80%93Strong_string_synthesis)

Use it like in this Videos:
https://youtu.be/2OnzblGvRDk

https://youtu.be/PU5awXvGM0Q

GUI PR: mruby-zest/mruby-zest-build#121

### DIFF
--- a/src/DSP/CombFilter.cpp
+++ b/src/DSP/CombFilter.cpp
@@ -13,9 +13,14 @@
 
 namespace zyn{
 
+
 CombFilter::CombFilter(Allocator *alloc, unsigned char Ftype, float Ffreq, float Fq,
     unsigned int srate, int bufsize)
-    :Filter(srate, bufsize), gain(1.0f), q(Fq), type(Ftype), Plpf(127), Phpf(0), 
+    :CombFilter(alloc, Ftype, Ffreq, Fq, srate, bufsize, 127, 0){}
+
+CombFilter::CombFilter(Allocator *alloc, unsigned char Ftype, float Ffreq, float Fq,
+    unsigned int srate, int bufsize, unsigned char Plpf_, unsigned char Phpf_)
+    :Filter(srate, bufsize), gain(1.0f), q(Fq), type(Ftype), Plpf(127), Phpf(0),
     lpf(nullptr), hpf(nullptr), memory(*alloc)
 {
     //worst case: looking back from smps[0] at 25Hz using higher order interpolation
@@ -27,6 +32,8 @@ CombFilter::CombFilter(Allocator *alloc, unsigned char Ftype, float Ffreq, float
 
     setfreq_and_q(Ffreq, q);
     settype(type);
+    setlpf(Plpf_);
+    sethpf(Phpf_);
 }
 
 CombFilter::~CombFilter(void)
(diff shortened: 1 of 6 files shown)


src/DSP/CombFilter.cpp
```diff
@@ -31,6 +33,13 @@ CombFilter::~CombFilter(void)
 {
     memory.dealloc(input);
     memory.dealloc(output);
+
+    if(Plpf != 127) { //LowPass
+        memory.dealloc(lpf);
+    }
+    if(Phpf != 0) { //HighPass
+        memory.dealloc(hpf);
+    }
```
fundamental, line 42: Minor, but IIRC dealloc will be a no-op if the input value is `nullptr` and it will replace the old pointer with `nullptr` when actually performing a dealloc, so the conditions shouldn't be needed?

src/DSP/Filter.cpp
```diff
@@ -59,8 +59,13 @@ Filter *Filter::generate(Allocator &memory, const FilterParams *pars,
...
         case 4:
-            filter = memory.alloc<CombFilter>(&memory, Ftype, 1000.0f, pars->getq(), srate, bufsize);
-            filter->setgain(pars->getgain());
+            {
+            CombFilter* comb = memory.alloc<CombFilter>(&memory, Ftype, 1000.0f, pars->getq(), srate, bufsize);
+            comb->setgain(pars->getgain());
+            comb->sethpf(pars->Phpf);
+            comb->setlpf(pars->Plpf);
+            filter = comb;
+            }
```
fundamental, line 68: Would we want to have a second constructor for CombFilter that accepts gain/hpf/lpf? That would remove the need for a variable scoped to a single case statement.

fundamental:
At least my reading only shows one typo (hpl->hpf) and then a few minor stylistic comments. Let me know if you're happy with the sound this produces and I'll do a couple of local demos and a quick sanity test (I'm assuming the CI failure is one of the flaky tests).

src/Synth/ModFilter.cpp
```diff
@@ -77,6 +77,7 @@ void ModFilter::update(float relfreq, float relq)
         baseFreq = pars.getfreq();
         baseQ    = pars.getq();
         tracking = pars.getfreqtracking(noteFreq);
+
```
JohannesLorenz, line 80: Do you really want to commit this?

src/DSP/Filter.cpp
friedolino78: ok

src/DSP/CombFilter.cpp
friedolino78: tried it without. seems to work

src/DSP/CombFilter.cpp
```diff
@@ -25,12 +32,21 @@ CombFilter::CombFilter(Allocator *alloc, unsigned char Ftype, float Ffreq, float
...
+    sethpf(Phpf_);
 }
 
 CombFilter::~CombFilter(void)
 {
     memory.dealloc(input);
     memory.dealloc(output);
+
+
+        memory.dealloc(lpf);
```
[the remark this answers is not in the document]
friedolino78: fixed

friedolino78:
solved via https://github.com/zynaddsubfx/zynaddsubfx/pull/337

Issue #309, opened 2024-12-16
Remaining issues from Reverse PR (#255)


JohannesLorenz:
* Make FLTK GUI
* Make standalone Lv2 effect
* Make standalone Lv2 effect for sympathetic

PR #326, opened 2025-03-16
add a Reverse lv2 effect

friedolino78:
https://github.com/zynaddsubfx/zynaddsubfx/issues/309

### DIFF
--- a/src/Effects/EffectMgr.cpp
+++ b/src/Effects/EffectMgr.cpp
@@ -316,6 +316,7 @@ void EffectMgr::changeeffectrt(int _nefx, bool avoidSmash)
     preset = 0;
     memset(efxoutl, 0, synth.bufferbytes);
     memset(efxoutr, 0, synth.bufferbytes);
+    if(sync) sync->detach(efx);
     memory.dealloc(efx);
 
     int new_loc = (_nefx == 8) ? dynfilter_0 : in_effect;
--- a/src/Misc/Time.h
+++ b/src/Misc/Time.h
@@ -18,18 +18,8 @@ namespace zyn {
 class AbsTime
 {
     public:
-        AbsTime(const SYNTH_T &synth)
-            :tempo(120),
-            hostSamples(0),
-            bar(0),
-            beat(0),
-            tick(0.0f),
-            bpm(120.0f),
-            ppq(1920.0f),
-            frames(0),
-            samplingInterval(synth.dt()),
-            buffersize(synth.buffersize) {};
-        AbsTime(const int dt_, const int buffersize_)
+        AbsTime(const SYNTH_T &synth) : AbsTime(synth.dt(), synth.buffersize) {}
+        AbsTime(const float dt_, const int buffersize_)
             :tempo(120),
             hostSamples(0),
             bar(0),
@@ -59,8 +49,8 @@ class AbsTime
         int   samplerate() const {return buffersize / samplingInterval;}
     private:
         int64_t frames;
-        float samplingInterval;
-        int buffersize;
+        const float samplingInterval;
+        const int buffersize;
 
 };
 
--- a/src/Misc/Time.h
+++ b/src/Misc/Time.h
@@ -27,7 +27,19 @@ class AbsTime
             bpm(120.0f),
             ppq(1920.0f),
             frames(0),
-            s(synth) {};
+            samplingInterval(synth.dt()),
+            buffersize(synth.buffersize) {};
+        AbsTime(const int dt_, const int buffersize_)
+            :tempo(120),
+            hostSamples(0),
+            bar(0),
+            beat(0),
+            tick(0.0f),
+            bpm(120.0f),
+            ppq(1920.0f),
+            frames(0),
+            samplingInterval(dt_),
+            buffersize(buffersize_) {};
         void operator++(){++frames;};
         void operator++(int){frames++;};
         int64_t time() const {return frames;};
@@ -41,13 +53,14 @@ class AbsTime
         float bpm;
         float ppq;
         bool playing;
-        float dt() const { return s.dt(); }
-        float framesPerSec() const { return 1/s.dt();}
-        int   samplesPerFrame() const {return s.buffersize;}
-        int   samplerate() const {return s.buffersize / s.dt();}
+        float dt() const { return samplingInterval; }
+        float framesPerSec() const { return 1.0f/samplingInterval;}
+        int   samplesPerFrame() const {return buffersize;}
+        int   samplerate() const {return buffersize / samplingInterval;}
     private:
         int64_t frames;
-        const SYNTH_T &s;
+        float samplingInterval;
+        int buffersize;
 
 };
 
--- a/src/Plugin/AbstractFX.hpp
+++ b/src/Plugin/AbstractFX.hpp
@@ -20,8 +20,8 @@
 
 */
 
-//~ #ifndef ZYNADDSUBFX_ABSTRACTFX_HPP_INCLUDED
-//~ #define ZYNADDSUBFX_ABSTRACTFX_HPP_INCLUDED
+#ifndef ZYNADDSUBFX_ABSTRACTFX_HPP_INCLUDED
+#define ZYNADDSUBFX_ABSTRACTFX_HPP_INCLUDED
 
 // DPF includes
 #include "DistrhoPlugin.hpp"
@@ -185,15 +185,15 @@ protected:
         else
             multiply(outputs[1], 0.5f, frames);
 
-    if(timePosition.bpm) {
+    if(timePosition.bbt.valid) {
         time.hostSamples = frames;
-        time.bar = timePosition.bar;
-        time.beat = timePosition.beat;
-        time.tick = timePosition.tick;
-        time.beatsPerBar = timePosition.beatsPerBar;
-        time.tempo = timePosition.bpm;
-        time.bpm = timePosition.bpm;
-        time.ppq = timePosition.PPQ;
+        time.bar = timePosition.bbt.bar;
+        time.beat = timePosition.bbt.beat;
+        time.tick = timePosition.bbt.tick;
+        time.beatsPerBar = timePosition.bbt.beatsPerBar;
+        time.tempo = timePosition.bbt.beatsPerMinute;
+        time.bpm = timePosition.bbt.beatsPerMinute;
+        time.ppq = timePosition.bbt.ticksPerBeat;
         time.playing = timePosition.playing;
     }
     else {
@@ -315,4 +315,4 @@ private:
     DISTRHO_DECLARE_NON_COPY_CLASS(AbstractPluginFX)
 };
 
-//~ #endif // ZYNADDSUBFX_ABSTRACTFX_HPP_INCLUDED
+#endif // ZYNADDSUBFX_ABSTRACTFX_HPP_INCLUDED
--- a/src/Plugin/AbstractFX.hpp
+++ b/src/Plugin/AbstractFX.hpp
@@ -73,7 +73,7 @@ public:
         delete filterpar;
     }
 
-    virtual void doReinit(const bool firstInit)
+    virtual void doReinit(bool firstInit)
     {
         // save current param values before recreating effect
         uchar params[paramCount];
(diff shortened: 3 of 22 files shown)


src/Plugin/Reverse/DistrhoPluginInfo.h
```diff
@@ -0,0 +1,30 @@
...
+
+#ifndef DISTRHO_PLUGIN_INFO_H_INCLUDED
+#define DISTRHO_PLUGIN_INFO_H_INCLUDED
+
+#define DISTRHO_PLUGIN_BRAND "ZynAddSubFX"
+#define DISTRHO_PLUGIN_NAME  "ZynReverse"
+#define DISTRHO_PLUGIN_URI   "http://zynaddsubfx.sourceforge.net/fx#Reverse"
+
+#define DISTRHO_PLUGIN_HAS_UI        0
+#define DISTRHO_PLUGIN_WANT_TIMEPOS  1
```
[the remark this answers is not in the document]
JohannesLorenz: I guess this comment can be resolved?

src/Plugin/AbstractFX.hpp
```diff
@@ -65,6 +74,38 @@ class AbstractPluginFX : public Plugin
...
+        effect = instantiateFX(pars);
+
+        if (firstInit)
+        {
+            effect->setpreset(0);
+        }
+        else
+        {
+            for (int i=0, count=static_cast<int>(paramCount); i<count; ++i)
+                effect->changepar(i+2, params[i]);
```
JohannesLorenz, line 102: This is old code, but I wonder: is `params` used uninitialized here if effect was `nullptr` above?

src/Plugin/Reverse/Reverse.cpp
```diff
@@ -0,0 +1,237 @@
...
+        return "Reverse";
+    }
+
+   /**
+      Get an extensive comment/description about the plugin.
+    */
+    const char* getDescription() const noexcept override
+    {
+        // TODO
+        return "";
```
JohannesLorenz, line 53: Maybe fill that out and remove the TODO?

src/Plugin/Reverse/Reverse.cpp
```diff
@@ -0,0 +1,237 @@
...
+   /* --------------------------------------------------------------------------------------------------------
+    * Init */
+
+   /**
+      Initialize the parameter @a index.
+      This function will be called once, shortly after the plugin is created.
+    */
+void initParameter(uint32_t index, Parameter& parameter) noexcept override
+{
+    parameter.hints = kParameterIsInteger | kParameterIsAutomable; // Alle Parameter automatisierbar
```
JohannesLorenz, line 133:
```suggestion
    parameter.hints = kParameterIsInteger | kParameterIsAutomable; // All parameters automatable
```

src/Plugin/Reverse/Reverse.cpp
```diff
@@ -0,0 +1,237 @@
...
+    parameter.unit  = "";
+    parameter.ranges.min = 0.0f;
+    parameter.ranges.max = 127.0f;
+    parameter.ranges.def = 64.0f; // Standardwert für die meisten Parameter
```
JohannesLorenz, line 137:
```suggestion
    parameter.ranges.def = 64.0f; // default value for most parameters
```

JohannesLorenz (commented):
Initial code review.

Follow-up code review and testing should still be done.

src/Effects/EffectMgr.cpp
```diff
@@ -316,6 +316,7 @@ void EffectMgr::changeeffectrt(int _nefx, bool avoidSmash)
     preset = 0;
     memset(efxoutl, 0, synth.bufferbytes);
     memset(efxoutr, 0, synth.bufferbytes);
+    if(sync) sync->detach(efx);
```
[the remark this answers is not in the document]
friedolino78: it is also called in EffectMgr::init

src/Misc/Time.h
```diff
@@ -27,7 +27,19 @@ class AbsTime
             bpm(120.0f),
             ppq(1920.0f),
             frames(0),
-            s(synth) {};
+            samplingInterval(synth.dt()),
```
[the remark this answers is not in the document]
friedolino78: no clue, too complicated for me. but if you send me code i can replace it.

src/Misc/Time.h
```diff
@@ -41,13 +53,14 @@ class AbsTime
...
-        int   samplesPerFrame() const {return s.buffersize;}
-        int   samplerate() const {return s.buffersize / s.dt();}
+        float dt() const { return samplingInterval; }
+        float framesPerSec() const { return 1.0f/samplingInterval;}
+        int   samplesPerFrame() const {return buffersize;}
+        int   samplerate() const {return buffersize / samplingInterval;}
     private:
         int64_t frames;
-        const SYNTH_T &s;
+        float samplingInterval;
```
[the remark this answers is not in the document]
friedolino78: ok

src/Plugin/AbstractFX.hpp
```diff
@@ -54,8 +56,15 @@ class AbstractPluginFX : public Plugin
...
-        doReinit(true);
+        
+        // set default tempo
+        time.tempo = 120;
+        time.bar = 0;
+        time.beat = 0;
+        time.tick = 0.0f;
+        time.bpm = 0.0f;
+
+        //~ doReinit(true);
```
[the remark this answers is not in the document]
friedolino78: no. i remove it.

src/Plugin/Reverse/Reverse.cpp
friedolino78: ok

src/Misc/Time.h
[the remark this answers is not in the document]
JohannesLorenz: Suggestion pushed as friedolino78/zynaddsubfx#6

src/Plugin/AbstractFX.hpp
JohannesLorenz: @fundamental do you have an idea here?

src/Misc/Time.h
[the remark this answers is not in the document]
friedolino78: it is now

src/Plugin/AbstractFX.hpp
```diff
@@ -65,6 +74,38 @@ class AbstractPluginFX : public Plugin
         delete effect;
         delete filterpar;
     }
+    
+    virtual void doReinit(const bool firstInit)
+    {
+        // save current param values before recreating effect
+        uchar params[paramCount];
```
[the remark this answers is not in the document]
fundamental: Perhaps historical at this point, but I recall one user had librtosc compiling via MSVC, but for the zyn build scripts it's all gcc/clang including windows support. Originally this was due to MSVC quirks around things like lambdas in initializer lists (aka the whole rtosc::Ports usage within zyn), so it shouldn't be a problem for the existing build pipeline.

src/Plugin/AbstractFX.hpp
```diff
@@ -55,7 +57,12 @@ class AbstractPluginFX : public Plugin
         std::memset(efxoutl, 0, sizeof(float)*bufferSize);
         std::memset(efxoutr, 0, sizeof(float)*bufferSize);
 
-        doReinit(true);
```
fundamental, line 58: Are we sure this line should be removed?

src/Plugin/AbstractFX.hpp
```diff
@@ -178,6 +222,20 @@ class AbstractPluginFX : public Plugin
         else
             multiply(outputs[1], 0.5f, frames);
 
+        if(timePosition.bbt.valid) {
```
fundamental, line 225: General question: is timePosition.bbt.valid true for all frames in a host that has valid time position information or is there the risk that it's only valid for some frames?

src/Plugin/AbstractFX.hpp
fundamental: I see in the other changes doReinit() is called in the implementation classes, so I guess this is fine. It might have been a little simpler to add a dummy doReinit() to Reverse.h, but everything checks out.

src/Plugin/AbstractFX.hpp
fundamental: Hm, a second reading through it looks like effect _should_ only be a nullptr if and only if it's the firstInit. Maybe as a guard make firstInit non-const and put an else statement at line 87 which forces it to true if the effect was null?

src/Plugin/Reverse/Reverse.cpp
```diff
@@ -0,0 +1,236 @@
...
+    /**
+      Get the current value of a parameter.
+      The host may call this function from any context, including realtime processing.
+    */
+    float getParameterValue(uint32_t index) const override
+    {
+        switch(index)
+        {
+            case 4:
+                return AbstractPluginFX::getParameterValue(index)-2;
```
fundamental, line 79: Looks like a bug. Were you targeting `getParameterValue(index-offset)`?

src/Plugin/Reverse/Reverse.cpp
```diff
@@ -0,0 +1,236 @@
...
+      Change a parameter value.
+      The host may call this function from any context, including realtime processing.
+      When a parameter is marked as automable, you must ensure no non-realtime operations are performed.
+      @note This function will only be called for parameter inputs.
+    */
+    void setParameterValue(uint32_t index, float value) override
+    {
+        if(index==4)
+        {
+            AbstractPluginFX::setParameterValue(index, value+2);
```
fundamental, line 104: Same as before, should the offset be applied to the index?

src/Plugin/Reverse/Reverse.cpp
```diff
@@ -0,0 +1,236 @@
...
+#include "Effects/Reverse.h"
+
+/* ------------------------------------------------------------------------------------------------------------
+ * Reverse plugin class */
+
+class ReversePlugin : public AbstractPluginFX<zyn::Reverse>
+{
+public:
+    ReversePlugin()
+        : AbstractPluginFX(9, 5) {}
```
fundamental, line 27: Missing initializer for numerator/denominator fields?

src/Plugin/Reverse/Reverse.cpp
```diff
@@ -0,0 +1,236 @@
...
+            AbstractPluginFX::setParameterValue(index, value+2);
+            return;
+        }
+
+        if(index==5 || index==6)
+        {
+            if(index==5) numerator = value;
+            if(index==6) denominator = value;
+            if (numerator&&denominator)
+                        effect->speedfactor = (float)denominator / (4.0f *(float)numerator);
```
fundamental, line 113: minor - inconsistent formatting

fundamental:
Looks reasonable overall with this review catching a possible set of uninitialized member variables and a possible bug with an indexing offset.

src/Plugin/Reverse/Reverse.cpp
friedolino78: fixed

src/Plugin/Reverse/Reverse.cpp
friedolino78: hmm, maybe it's not a bug?

friedolino78: it's a mapping of the 4 sync modes to the two supported in plugin mode
sync to noteOn/Off is not available in a plugin.

src/Plugin/AbstractFX.hpp
friedolino78: done

src/Plugin/Reverse/Reverse.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+Plugin* createPlugin()
+{
+    Plugin* plugin = new ReversePlugin();
+    ReversePlugin* fxPlugin = dynamic_cast<ReversePlugin*>(plugin);
+    if (fxPlugin)
+    {
+        fxPlugin->doReinit(true);
+        fxPlugin->setSpeedfactor(1.0f);
+    }
```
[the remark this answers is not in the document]
friedolino78: ok

JohannesLorenz:
> Good to merge?
> 
> @JohannesLorenz you mentioned something about the macos CI, would that be something that would address the `Assertion failed: (o.padpars == NULL), function operator(), file Part.cpp, line 269.` error in `saveOscAllFiles`?

That test fails on master, too. No need to fix it here.

I'll merge.

PR #336, opened 2025-07-12
Wave shape refactor

friedolino78:
suggested refactoring by perplexity.ai
some feature completion

### DIFF
--- a/src/Misc/WaveShapeSmps.cpp
+++ b/src/Misc/WaveShapeSmps.cpp
@@ -85,24 +85,25 @@ void processArctangent(int n, float* smps, float ws, float offs) {
     }
 }
 
-void processAsymmetric(int n, float* smps, float ws) {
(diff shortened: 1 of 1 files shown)


friedolino78:
I would say let's finish this first. 
after that I rebase the 210 and add some other new candidates (soft fold back). Then we find out which are really worth adding.

fundamental:
Running zyn and listening to the output it doesn't see like there are any glitches, but I will say two things cropped up:
- tanh is much louder than other distortions (given a sawtooth input), but that likely is an issue that predates this PR
- asym1/asym2 get \*very\* loud with offset around a value of 20-30 (sawtooth input again). Is there some way to compensate for that or is it something we're pretty much stuck with?

JohannesLorenz:
Should this go into Rework, or is fundamental's last issue fixed already (then "In Review" is already fine)?

PR #342, opened 2025-08-01
Fmaa sync f max

friedolino78:
WIP Please check last commit


fundamental:
I'm guessing the same answer as the other anti-aliasing PR, but should this yield differeing values in the unit tests with AA turned off?

PR #344, opened 2025-08-17
compensate lpf delay in comb filter

friedolino78:
small fix that keeps the comb filter resonance frequency constant when using the feedback LPF


fundamental:
Merged :+1:

PR #346, opened 2025-08-30
fix oob read in samplelerp

friedolino78:


fundamental:
Code seems good to me and CI is passing :+1:

PR #350, opened 2025-10-22
Fix missing FL include

JohannesLorenz:
This fixes the following error which seems to be due to newer FL versions:

```
  src/UI/MasterUI.cxx:2381:3: error: 'fl_open_display' was not declared in this scope
```

This also fixes 2 deprecation warnings.


fundamental:
Thanks for the fix :+1: Merged.

PR #351, opened 2025-10-26
Bump C++ version to C++17

JohannesLorenz:


fundamental:
Based on previous comments this should be good to merge. Any other changes or are we good to go?

PR #358, opened 2025-12-17
Chorus: Ensemble: reduce triple mode amplitude on left channel 

soyersoyer:
In triple mode, the right channel has an amplitude reduction,
but the left has not.

Apply the same 0.85 multiplier to it.


fundamental:
Good catch. :+1:

PR #362, opened 2026-01-17
MasterUI.fl: Fix `fl_open_display`/`HAS_X11`

JohannesLorenz:
This fixes multiple issues:

* fl_open_display on Windows is not available before FL 1.4.0 (this can be verified by looking at FLTK's sources: `src/Fl_win32.cxx`)
* Since `fl_open_display` is not called if `HAS_X11` is set, the `FL/platform.H` should not have the `HAS_X11` restriction either
* Since X11/xpm.h does require `HAS_X11`, do require it for the pixmap handling code, too

Fixup of
* e9a1b628b0fdcf2c1f2c88a8d3e9c964fb3bb607
* 3d73a99acfec68dfcca79d0295576dead8a40479

### DIFF
--- a/src/UI/MasterUI.fl
+++ b/src/UI/MasterUI.fl
@@ -97,7 +97,7 @@ decl {\#if !defined(PLUGINVERSION) && HAS_X11
 \#endif} {private local
 }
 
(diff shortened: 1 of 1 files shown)


fundamental:
Everything builds locally and the window spawns as expected :+1: Merged

PR #365, opened 2026-01-20
Rename variables hiding others

JohannesLorenz:
This helps with an MSVC warning and makes the code more clear.

Note: If this PR feels wrong, I can also disable the warning in MSVC, let me know what you prefer.

After this PR, all other PRs depend on the C++17 PR, which is next.


fundamental:
I see no typos, so I'll let CI verify and merge :+1:

PR #368, opened 2026-01-31
Fix warnings about unused variables

JohannesLorenz:


fundamental:
Everything looks good to me. Merged :+1:

PR #369, opened 2026-01-31
Fix/modernize CMakeLists

JohannesLorenz:
Changes, in order:

1. Fix LibGit2 warning
2. Use lists for handling CMake warnings (and different names of those lists to avoid issues in CMake cache) and use `add_compile_options`
3. Remove redundant "-Wall" specification
4. Suppress warnings about "unsafe-buffer-usage" and "implicit-int-float-conversion"
5. Print GUI status message
6. Fix warnings from DPF headers showing up in zyn build


fundamental:
I'll make a mental note that the SYSTEM include dir flag might mean we need a make clean when doing the DPF upgrade, but everything else looks sane. Merging :+1:

PR #371, opened 2026-02-06
Fix (implicit) type conversions

JohannesLorenz:
This commit mostly turns all integer parameters that behave boolean into real `bool` parameters.

In order to make sure that none of these parameters ever gets assigned with or compared to an integer, a temporary `SafeBool` class was used for verification, which also showed some places where integers were used and now were replaced by bools.

Aside from the parameters, there are a few minor other type conversion fixes.

### DIFF
--- a/src/Synth/Resonance.h
+++ b/src/Synth/Resonance.h
@@ -54,7 +54,7 @@ class Resonance:public Presets
         unsigned char Prespoints[N_RES_POINTS]; //how many points define the resonance function
         unsigned char PmaxdB;       //how many dB the signal may be amplified
         unsigned char Pcenterfreq, Poctavesfreq; //the center frequency of the res. func., and the number of octaves
-        bool Pprotectthefundamental;   //the fundamental (1-st harmonic) is not damped, even it resonance function is low
+        bool Pprotectthefundamental;   //the fundamental (1-st harmonic) is not damped, even if resonance function is low
 
         //controllers
         float ctlcenter; //center frequency(relative)


JohannesLorenz:
@fundamental This PR is quite large. Let me know if you want me to split it.

src/Misc/MiddleWare.cpp
```diff
@@ -2211,7 +2211,7 @@ void MiddleWareImpl::broadcastToRemote(const char *rtmsg)
 
 void MiddleWareImpl::sendToRemote(const char *rtmsg, std::string dest)
 {
-    if(!rtmsg || rtmsg[0] != '/' || !rtosc_message_length(rtmsg, -1)) {
+    if(!rtmsg || rtmsg[0] != '/' || !rtosc_message_length(rtmsg, (std::numeric_limits<size_t>::max)())) {
```
fundamental: While I'm not suggesting any change it does feel less clean than the old approach (even with the wonky conversion).

src/Params/ADnoteParameters.cpp
```diff
@@ -1002,8 +1002,8 @@ void ADnoteGlobalParam::getfromXML(XMLwrapper& xml)
         PPunchStretch  = xml.getpar127("punch_stretch", PPunchStretch);
         PPunchVelocitySensing = xml.getpar127("punch_velocity_sensing",
                                                PPunchVelocitySensing);
-        Hrandgrouping = xml.getpar127("harmonic_randomness_grouping",
-                                       Hrandgrouping);
+        Hrandgrouping = (bool) xml.getpar("harmonic_randomness_grouping",
```
fundamental: Changing from getpar127->getpar might be different under the hood. I'll at least try to circle back on this change.

src/Params/Controller.cpp
```diff
@@ -503,15 +503,15 @@ void Controller::getfromXML(XMLwrapper& xml)
...
+                                                   portamento.pitchthreshtype, 0, 1);
+    portamento.portamento = (bool) xml.getpar("portamento_portamento",
+                                              portamento.portamento, 0, 1);
     portamento.updowntimestretch = xml.getpar127(
         "portamento_updowntimestretch",
         portamento.updowntimestretch);
-    portamento.proportional = xml.getpar127("portamento_proportional",
-                                             portamento.proportional);
+    portamento.proportional = (bool) xml.getpar("portamento_proportional",
+                                                portamento.proportional, 0, 1);
```
fundamental: Another note to self with the getpar/getpar127 variation.

src/Params/PADnoteParameters.cpp
```diff
@@ -1220,7 +1220,7 @@ void PADnoteParameters::getfromXML(XMLwrapper& xml)
     }
 
     if(xml.enterbranch("FREQUENCY_PARAMETERS")) {
-        Pfixedfreq    = xml.getpar127("fixed_freq", Pfixedfreq);
+        Pfixedfreq    = (bool)xml.getpar("fixed_freq", Pfixedfreq, 0, 1);
```
fundamental: I'm suspecting these are fine based on how many I've seen, but another note to self.

fundamental: Seems sane on review.

src/Params/Controller.cpp
fundamental: Seems sane on the second pass.

src/Params/ADnoteParameters.cpp
fundamental: Seems sane on the second pass.

fundamental:
Looks good. To avoid a bunch of "Applied suggested change" commits I'm going to squash and merge :+1:

PR #374, opened 2026-02-17
CombFilterBank: set cutoff before samplerate

soyersoyer:
Value_Smoothing_Filter::sample_rate() function uses the cutoff, so it must be set beforehand.


friedolino78:
> This changes the sound, I don't know if it's desirable or not. But if not, it should be deleted or set to the default value.

oopsie - clicked before reading ;-)
are you on our discord?

soyersoyer:
No, but does one exist? :) I can't find it.

PR #376, opened 2026-02-24
add missing "delete sync" (valgrind)

friedolino78:


fundamental:
Thanks for fixing the leak :+1:

PR #383, opened 2026-03-17
Bump C++ version to C++20

JohannesLorenz:
@fundamental After some discussions with AI about #372 , it would be better if I use `std::thread*` instead of `std::thread` *everywhere* in that PR (all classes that I changed) because `std::thread::joinable()` is not reliable, and I don't want to introduce new issues. However, using pointers/`new` just for that does not seem idiomatic for me, so the only real alternative is C++20 to me, since it has guaranteed lock-free `std::atomic_flag::test()`. So, I pushed this branch. Please take a look at it, and please also compile it with your system, which seems to be a good testing candidate for an old Linux.

---

We require this change in order to use `std::atomic_flag::test()`.

Changes in this PR:

- Raise C++ version in CMakeLists.txt
- Fix `identity` clashing with
- Fix (probably) unrelated compiler warning in ADnoteUI.fl

## Summary by CodeRabbit

* **Chores**
  * Updated project to use C++20 standard for improved language features and compatibility.
  * Implemented code modernization improvements including parameter annotations and lambda optimizations.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,4 +1,4 @@
-set (CMAKE_CXX_STANDARD 17)
+set (CMAKE_CXX_STANDARD 20)
 
 #checking include/library paths
 message(STATUS "Checking Include Path" $ENV{CMAKE_INCLUDE_PATH} ${CMAKE_INCLUDE_PATH})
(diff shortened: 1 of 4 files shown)


fundamental:
Compiles and runs with fltk and zest locally. All changes seem reasonable to me. Merged :+1:

PR #398, opened 2026-07-10
Add tRelease to fix watch discontinuity on forced release

friedolino78:
When forced release triggers, t is reset to 0.0 to start the release segment. Save the previous t as tRelease and add it to the watch time offset, so the watch display shows a continuous timeline without jumping backwards.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved envelope timing reports when a forced release occurs.
  * Ensured monitored output positions remain accurate across forced releases.


friedolino78:
Rebased on latest upstream/master, builds clean. Ready for review.

PR #402, opened 2026-07-26
Windows ci test

JohannesLorenz:


JohannesLorenz:
@copilot Please do minimal changes, restricted to only what I ask you for:
1. The dependencies from dumpbin/objdump should be redirected into text files that are later archived as artifacts
2. There should be only one step which solely calls `matrix.config.dump_dependencies`, which does either the Windows or the MinGw step (or for Linux/Mac does something similar to show the dependencies). You can use a `matrix.config.dump_dependencies_shell` if you need different shells for that common step.

In the end, there should be one text file archived for each of the operating systems.

JohannesLorenz:
@copilot That did not work, the build failed.

Also, why did you complicate the code using vswhere and usebackq delims= ? Why make it so complicated?

JohannesLorenz:
@copilot Please fix the build error. Are you even able to read what the build error is? Because it seems you ignore it.

